PR #3, opened 2022-11-09
Discourse API

jdkandersson:
Adds an API for interactions with discourse.


weiiwang01:
Can you add bandit static check to the tox configuration file? `operator-workflows` will fail without that.

jdkandersson:
> Can you add bandit static check to the tox configuration file? `operator-workflows` will fail without that.

@weiiwang01 just to help me understand, `tox -e lint` currently includes `bandit` as well, is that what is needed? Or would it need to be a separate command?

weiiwang01:
> > Can you add bandit static check to the tox configuration file? `operator-workflows` will fail without that.
> 
> @weiiwang01 just to help me understand, `tox -e lint` currently includes `bandit` as well, is that what is needed? Or would it need to be a separate command?

Yes, bandit needs to be put in a separate tox env called `static` (well, you can put anything static check related in `static` but the `static` testenv has to exist). Here's an example:
```
[testenv:static]
description = Run static analysis tests
deps =
    bandit
    toml
    -r{toxinidir}/requirements.txt
commands =
    bandit -c {toxinidir}/pyproject.toml -r {[vars]src_path} {[vars]tst_path}
```

weiiwang01:
> > > Can you add bandit static check to the tox configuration file? `operator-workflows` will fail without that.
> > 
> > 
> > @weiiwang01 just to help me understand, `tox -e lint` currently includes `bandit` as well, is that what is needed? Or would it need to be a separate command?
> 
> The `operator-workflows` requires certain testenv exists and running by default, and `static` is one of them. So, bandit needs to be put in a separate tox env called `static` (well, you can put anything static check related in `static` but the `static` testenv has to exist). Here's an example:
> 
> ```
> [testenv:static]
> description = Run static analysis tests
> deps =
>     bandit
>     toml
>     -r{toxinidir}/requirements.txt
> commands =
>     bandit -c {toxinidir}/pyproject.toml -r {[vars]src_path} {[vars]tst_path}
> ```

`operator-workflows` uses testenv name to extract different test results, so these four testenv `unit`, `lint`, `static` and `coverage-report` must exist and be include in tox envlist.

arturo-seijas (commented):
The unit tests are failing because `coverage-report`is not among the tox default targets. I'm wondering if it would be better to make the pipeline pass in this scenario or if there's a way to provide useful feedback and require the execution of this step.

.github/workflows/integration_test.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+on:
+  pull_request:
+
+jobs:
+  integration-tests:
+    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@v1.0.0
+    secrets: inherit
+    with:
+      series: jammy
```
arturo-seijas, line 11: This is what's causing the integration tests to fail. The parameter format is an enquoted json list: `'["jammy"]'`, unfortunately strings are the only way to pass complex arguments through steps.
On the other hand, I don't think you need this. The base image to run your tests will be the one defined in the `build-on` field in `charmcraft.yaml`. This is aimed at running the integration tests once per specified `series` while addding a `--series` command line parameter (as in https://github.com/jenkinsci/jenkins-charm/blob/master/.github/workflows/ci.yaml#L41)

tests/integration/conftest.py
```diff
@@ -0,0 +1,284 @@
...
+    async def get_discourse_status():
+        """Get the status of discourse."""
+        return (await ops_test.model.get_status())["applications"]["discourse-k8s"].status[
+            "status"
+        ]
+
+    for _ in range(120):
+        if await get_discourse_status() != "waiting":
+            break
+        await asyncio.sleep(10)
```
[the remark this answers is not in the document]
jdkandersson: I tried that, there is a wait_for_idle call just above this, the problem is that the discourse charm will enter the idle and active state whilst still waiting for discourse to be setup in the container, which takes like 5 to 10 minutes. The tests depend on that finishing. I'll add more details to the comment including why wait for idle and the active state doesn't work here.

.github/workflows/integration_test.yaml
```diff
@@ -0,0 +1,11 @@
+name: Integration Tests
+
+on:
+  pull_request:
+
+jobs:
+  integration-tests:
+    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@enable-ingress-integration-tests
```
arturo-seijas, line 8: This is not pointing to a release. I assume you'll want to change it before merging

src/discourse.py
```diff
@@ -0,0 +1,288 @@
...
+        Args:
+            url: The URL to the topic. Assume it includes the slug and id of the topic as the last
+                2 elements of the url.
+
+        Returns:
+            The content of the first post in the topic.
+
+        """
+        first_post = self._retrieve_topic_first_post(url=url)
+        return first_post["cooked"].removeprefix("<p>").removesuffix("</p>")
```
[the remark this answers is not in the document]
jdkandersson: You are right, the stripping of the paragraph element should be removed on the next commit. This will be revisited when looking at how this works for markdown posted on discourse, it potentially means that documentation will always be updated rather than only on change which should be fine

src/discourse.py
```diff
@@ -0,0 +1,288 @@
...
+            The first post from the topic.
+
+        """
+        topic_info = self._retrieve_topic_info_from_url(url=url)
+        try:
+            topic = self.client.topic(slug=topic_info.slug, topic_id=topic_info.id_)
+        except pydiscourse.exceptions.DiscourseError as discourse_error:
+            raise DiscourseError from discourse_error
+
+        first_post = next(
```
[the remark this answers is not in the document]
jdkandersson: Thank you for this, yes there should be more checks on the return value from discourse. The next commits should include improved error handling and unit tests to check for these.

src/discourse.py
```diff
@@ -0,0 +1,288 @@
...
+            url: The URL to the topic.
+            content: The content for the first post in the topic.
+
+        """
+        first_post = self._retrieve_topic_first_post(url=url)
+        try:
+            self.client.update_post(
+                post_id=first_post["id"],
+                content=content,
+                edit_reason="Charm documentation updated",
```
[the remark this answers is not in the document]
jdkandersson: Good idea, should be included in next commit

tests/integration/conftest.py
```diff
@@ -0,0 +1,284 @@
...
+
+@pytest_asyncio.fixture(scope="module")
+async def discourse_admin_credentials(ops_test: OpsTest, discourse_unit_name: str):
+    """Get the admin credentials for discourse."""
+    return await create_discourse_account(
+        ops_test=ops_test, unit=discourse_unit_name, email="admin@foo.internal", admin=True
+    )
+
+
+# Making this depend on discourse_admin_credentials to ensure an admin user gets created
```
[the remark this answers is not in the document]
jdkandersson: Thank you, fixed. I will leave the comments just to clarify why the fixture is needed even though it isn't used

.github/workflows/integration_test.yaml
jdkandersson: Yes, just need that PR to enable ingress to be merged 😄

weiiwang01 (approved):

.github/workflows/integration_test.yaml
arturo-seijas: It has been merged :)

pyproject.toml
```diff
@@ -11,7 +11,7 @@ skips = ["*/*test.py", "*/test_*.py"]
 branch = true
 
 [tool.coverage.report]
-fail_under = 85
+fail_under = 100
```
merkata, line 14: bold move :grin:

.github/workflows/integration_test.yaml
jdkandersson: Thank you, switched this to point at main now

weiiwang01 (approved):

yhaliaw (approved):

PR #4, opened 2022-11-14
Retrieve or create the documentation index file

jdkandersson:
In this version, support for retrieving or creating the index file has been added


src/run.py
```diff
@@ -0,0 +1,138 @@
...
+        content = _read_docs_index(base_path=base_path)
+
+        try:
+            index_url = server_client.create_topic(
+                title=f"{name_value.replace('-', ' ').title()} Documentation Overview",
+                content=content,
+            )
+        except DiscourseError as exc:
+            raise ServerError("Index page creation failed") from exc
+    elif docs_key not in metadata and not create_if_not_exists:
```
gregory-schiano, line 126: I tend to always put the error case first in my logic, and then follow the "happy case"
It'll allow to factor a bit the code by using "_get_key" once.
Just a suggestion

src/run.py
```diff
@@ -0,0 +1,138 @@
...
+        raise InputError(
+            f"'{docs_key!r}' not defined in metadata.yaml and 'create_if_not_exists' false, "
+            f"{metadata=!r}"
+        )
+    else:
+        index_url = _get_key(metadata=metadata, key=docs_key)
```
gregory-schiano, line 132: I'm confused by the naming of the variable, in the `if` index_url comes out for `create_topic` and `_get_key` is affected to `name_value`
Same for `content` variable

tests/unit/test_run.py
```diff
@@ -0,0 +1,309 @@
...
+    arrange: given directory with metadata.yaml with docs defined and discourse client that
+        returns the index page content
+    act: when retrieve_or_create_index is called with that directory
+    assert: then retrieve topic is called with the docs key value and the content returned by the
+        client and docs key is returned.
+    """
+    url = "http://server/index-page"
+    content = "content 1"
+    metadata_yaml = tmp_path / "metadata.yaml"
+    metadata_yaml.write_text(f"docs: {url}", encoding="utf-8")
```
gregory-schiano, line 297: This is repeated multiple times, could be interesting to factor this in a method ?

weiiwang01 (dismissed):

src/run.py
```diff
@@ -0,0 +1,138 @@
...
+    Raises InputError if the file does not exist.
+
+    Args:
+        base_path: The starting path to look for the index content.
+
+    Returns:
+        The content of the index file.
+
+    """
+    if not (docs_folder := base_path / "docs").is_dir():
```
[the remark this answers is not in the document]
jdkandersson: Thanks, should be fixed in the next commit

src/discourse.py
```diff
@@ -329,3 +328,53 @@ def update_topic(
...
+        )
+    api_key = os.getenv("DISCOURSE_API_KEY")
+    if api_key is None:
+        raise InputError(
+            "The DISCOURSE_API_KEY environment variable is missing but is required to be able to "
+            "interact with the documentation server"
+        )
+
+    return Discourse(
+        base_path=f"https://{hostname}",
```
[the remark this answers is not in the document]
jdkandersson: This was deliberate for this function as it is essentially a factory method for creating the discourse client for execution in an opinionated way. The use case for using this function is for the action execution, which includes passing sensitive data to the discourse server, such as the API key. It is still possible to use the constructor for the `Discourse` class directly if it is required, the integration tests make use of this. The enforcement of https is to ensure the user doesn't accidentally expose their credentials. Does that make sense?

src/run.py
jdkandersson: Yes putting the elif as a separate condition first is better, that should be changed in the next commit. I don't follow being able to reduce the `_get_key` call, the two calls in this function are for different keys, one for the `docs` if that is defined in the metadata indicating that there is an existing documentation post and the other for the `name` key if the `docs` key is not in the metadata which indicates that there is no existing documentation post, in which case it is created. Or am I misunderstanding something?

src/run.py
jdkandersson: The next commit should add more documentation to explain this. Essentially they are two different cases, in one the index documentation page needs to be created and in the other it already exists so it just needs to be retrieved.

tests/unit/test_run.py
jdkandersson: Thanks, this should be fixed in the next commit

src/discourse.py
```diff
@@ -329,3 +328,53 @@ def update_topic(
...
+        raise InputError(
+            f"Invalid discourse_host input, it should not include the protocol, got {hostname=!r}"
+        )
+
+    if not isinstance(category_id, int):
+        raise InputError(
+            f"Invalid discourse_category_id input, it must be an integer, got {category_id=!r}"
+        )
+
+    api_username = os.getenv("DISCOURSE_API_USERNAME")
```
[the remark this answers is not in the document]
jdkandersson: Thanks fixed. I also changed this function to be more permissive with inputs and check those inputs.

src/discourse.py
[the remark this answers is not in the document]
gregory-schiano: Fine, as long as it's opiniated and therefore made on purpose :)

src/run.py
gregory-schiano: No misundertanding, I didn't read properly the code. Fine if we agree on the `elif`
If you don't mind I'll close the comment once the change is introduced ?

src/run.py
gregory-schiano: Ok, will see later, your explanation does make sense

arturo-seijas (approved):
LGTM

src/run.py
jdkandersson: It should already be done

amandahla (approved):

Issue #8, opened 2022-11-27
Create a PR with the updated `docs` metadata key


jdkandersson:
As an action user I want the action to create a PR adding in the URL to the documentation index after running if the index was created by the page so that I can update the metadata and ensure that the documentation is shown correctly on charmhub.

jdkandersson:
Won't do this

PR #11, opened 2022-11-29
Working alpha version of the action

jdkandersson:
- Add end to end code enabling the action
- Bug fixes


jdkandersson:
This is an example of it working: https://github.com/canonical/indico-operator/pull/188

.github/workflows/integration_test.yaml
```diff
@@ -21,24 +22,88 @@ jobs:
...
+          discourse_api_username: ${{ secrets.DISCOURSE_API_USERNAME }}
+          discourse_api_key: ${{ secrets.DISCOURSE_API_KEY }}
+      - name: Show pages
+        run: echo '${{ steps.selfTestCreate.outputs.urls_with_actions }}'
+      - name: Check create
+        run: |
+          ./discourse_check_cleanup.py --check-create --check-only '${{ steps.selfTestCreate.outputs.urls_with_actions }}' '${{ steps.selfTestCreate.outputs.discourse_config }}'
+      - name: Add docs key to metadata
+        run: |
+          echo "docs: ${{ steps.selfTestCreate.outputs.index_url }}" >> metadata.yaml
```
arturo-seijas, line 75: I have the feeling that this is a little obscure and TBH I don't like much the idea of modifying the code automatically. I guess we can't post to the same URL and upgrade the content each version, right?

action.yaml
```diff
@@ -2,11 +2,16 @@ name: Upload Charm Docs
 description: Upload charm documentation to charmhub
 author: David Andersson
 inputs:
-  create_new_topic:
+  delete_topics:
     description: |
-      Whether to create a new topic and commit it back to the repository if the
-      docs key in metadata.yaml either does not exist or is empty. Defaults to
-      not creating a new topic in which case an error is returned.
+      Whether to delete topics from the documentation server if they are no
```
arturo-seijas, line 7: What does "not longer needed" mean here?

action.yaml
```diff
@@ -2,11 +2,16 @@ name: Upload Charm Docs
...
+      longer needed.
+    default: true
+    required: false
+    type: boolean
+  draft_mode:
```
arturo-seijas, line 12: I suggest we rename this to `dry-run`. It seems clearer to me.

action.yaml
```diff
@@ -30,10 +35,12 @@ inputs:
     required: false
     type: integer
 outputs:
-  created_pages:
+  urls_with_actions:
```
arturo-seijas, line 38: How about `url_changes` instead?

action.yaml
```diff
@@ -30,10 +35,12 @@ inputs:
...
     description: |
-      A JSON map with the pages that have been created where the key is the
-      local path and the value is the URL to the created page.
+      A JSON map with the urls and the actions that have been taken against
+      them.
+  index_url:
+    description: The URL to the documentation index page.
```
arturo-seijas, line 43: Is this the base URL? Can we provide a clearer description?

discourse_check_cleanup.py
```diff
@@ -0,0 +1,306 @@
...
+
+    if not args.check_only:
+        for url in urls_with_actions.keys():
+            with contextlib.suppress(DiscourseError):
+                discourse.delete_topic(url=url)
+
+    sys.exit(0 if check_result else 1)
+
+
+def _check_url_count(
```
arturo-seijas, line 74: I'd expect all of this methods to throw exceptions instead of controlling "exceptional" flow with booleans

tox.ini
```diff
@@ -50,16 +52,17 @@ deps =
...
     pytest_asyncio
-    pydocstyle
+    pydocstyle>=2.10
+    ruff
 commands =
     pydocstyle {[vars]src_path}
     codespell {toxinidir}/. --skip {toxinidir}/.git --skip {toxinidir}/.tox \
       --skip {toxinidir}/build --skip {toxinidir}/lib --skip {toxinidir}/venv \
       --skip {toxinidir}/.mypy_cache --skip {toxinidir}/icon.svg
+    ruff {[vars]all_path}
```
arturo-seijas, line 65: Do we want to keep the old tooling that provides overlapping functionality?

src/action.py
```diff
@@ -0,0 +1,366 @@
...
+        discourse: A client to the documentation server.
+        draft_mode: If enabled, only log the action that would be taken.
+        name: The charm name to prefix to the created pages title.
+
+    Returns:
+        A report on the outcome of executing the action.
+    """
+    logging.info("draft mode: %s, action: %s", draft_mode, action)
+
+    if action.content is None:
```
arturo-seijas, line 47: This is a little difficult to follow

.github/workflows/integration_test.yaml
jdkandersson: This is an integration test for the action only, it has no effect on when the action is run by a user. There are plans for future features where the action creates a new PR to add the docs key back if needed: https://github.com/canonical/upload-charm-docs/issues/8

action.yaml
jdkandersson: So this is when, for example, a file is deleted in the repository which will usually result in the topic being deleted on discourse, unless `delete_topics` is set to false in which case the topic is left on discourse. This is based on feedback from the design document where people wanted the ability to have to action not delete topics on discourse

action.yaml
jdkandersson: Good suggestion, fixed

action.yaml
jdkandersson: Hmm, I'm not pleased with `urls_with_actions `. The thing is that `url_changes` could imply that only the urls where some changes have been made are returned. Instead, the action returns a list of all URLs to topics in a map with the action that was taken, which could be `noop` or `skip` which means that no changes were made. `create` and `update` are also included when changes have been made. Maybe `url_action_map`?

action.yaml
jdkandersson: It is the URL that should be set under the `docs` key in `metadata.yaml`. Maybe `The URL to the documentation index page on discourse. The docs key in metadata.yaml should be updated to this value after the first run.`?

discourse_check_cleanup.py
jdkandersson: I'll provide the reasoning for the current design and see what you think. This utility is mainly for the action integration tests and makes use of the action output to validate whether the test has passed. If any of the checks fail, it doesn't mean that the utility has encountered an error or exception since the utility doesn't depend on whether the checks pass just to execute the checks and report back whether they pass or fail, similar to how a linter might be written. In the case of a linter, an issue it finds isn't an exception for the linter, the linter is working fine, it just needs to inspect the code and report all the problems. The user of the linter can then decide whether the linter failing is exceptional flow. In this case, the GitHub workflow will use the return value of this utility in the same way it would use the 0 or 1 return value of a linter. Does that make sense?

tox.ini
jdkandersson: I was planning on using both the tools for a little while just to see if anything is caught by the overlapping tooling that isn't caught by `ruff`. This should definitely be removed in the near future, in one of the next demos I'll discuss `ruff` a bit and we can decide whether we adopt it as a team. So far I haven't found anything that `ruff` doesn't catch, except the copyright notice which is a useful tool we would give up

src/action.py
jdkandersson: I'll add a few comments, please let me know if that helps. Do note that some of these changes are part of the other PR, this PR will get much shorter once that one is merged

arturo-seijas (approved):
LGTM

jdkandersson:
Reopen so that the new reported change is against the current main

PR #14, opened 2022-12-01
Working alpha version

jdkandersson:
- Added end to end functionality
- Added end to end tests
- Bug fixes


arturo-seijas (approved):
LGTM

tests/integration/test___init__.py
```diff
@@ -0,0 +1,273 @@
...
+    # 12. with the nested directory removed
+    caplog.clear()
+    nested_dir.rmdir()
+
+    urls_with_actions = run(
+        base_path=tmp_path, discourse=discourse_api, dry_run=False, delete_pages=True
+    )
+
+    assert (urls := tuple(urls_with_actions)) == (doc_url, index_url)
+    assert_substrings_in_string(chain(urls, ("'delete'", "'update'", "'success'")), caplog.text)
```
gregory-schiano, line 243: Should we also assert that some reference of the old nested_dir_doc is present in the logs ?

tests/unit/test_discourse.py
```diff
@@ -112,6 +115,12 @@
             None,
             id="valid trailing /",
         ),
+        pytest.param(
+            "/t/slug/1",
```
gregory-schiano, line 119: Should use topic url prefix constant

tests/unit/test_discourse.py
```diff
@@ -408,6 +417,26 @@ def test_create_topic(monkeypatch: pytest.MonkeyPatch, base_path: str, discourse
     assert url == f"{base_path}/t/{topic_slug}/{topic_id}"
```
gregory-schiano, line 417: Should use topic url prefix constant

tests/unit/test_discourse.py
```diff
@@ -408,6 +417,26 @@ def test_create_topic(monkeypatch: pytest.MonkeyPatch, base_path: str, discourse
...
 
 
+def test_delete_topic(monkeypatch: pytest.MonkeyPatch, base_path: str, discourse: Discourse):
+    """
+    arrange: given a mocked discourse client
+    act: when delete_topic is called first without the base path and then with it
+    assert: then the url to the topic is returned.
+    """
+    mocked_client = mock.MagicMock(spec=pydiscourse.DiscourseClient)
+    url_path = "/t/slug/1"
```
gregory-schiano, line 427: Should use topic url prefix constant

tests/unit/test_discourse.py
```diff
@@ -475,16 +504,24 @@ def test_update_topic_discourse_error(
...
-    assert: then nothing is returned.
+    act: when given update_topic is called without base path and then with
+    assert: then topic url is returned.
     """
     mocked_client = mock.MagicMock(spec=pydiscourse.DiscourseClient)
     mocked_client.topic.return_value = {
         "post_stream": {"posts": [{"post_number": 1, "user_deleted": False, "id": 1}]}
     }
     monkeypatch.setattr(discourse, "_client", mocked_client)
+    url_path = "/t/slug/1"
```
gregory-schiano, line 515: Should use topic url prefix constant

tests/unit/test_discourse.py
```diff
@@ -584,11 +621,13 @@ def test_retrieve_topic_http_error(
...
+    mocked_session = mock.MagicMock(spec=requests.Session)
+    mock_get_requests_session.return_value = mocked_session
     mocked_response = mock.MagicMock(spec=requests.Response)
-    mocked_get.return_value = mocked_response
+    mocked_session.get.return_value = mocked_response
     mocked_response.raise_for_status.side_effect = requests.HTTPError
-    monkeypatch.setattr(requests, "get", mocked_get)
+    monkeypatch.setattr(discourse, "_get_requests_session", mock_get_requests_session)
 
     url = f"{base_path}/t/slug/1"
```
gregory-schiano, line 632: Should use topic url prefix constant

tests/unit/test_discourse.py
```diff
@@ -611,18 +650,43 @@ def test_retrieve_topic(monkeypatch: pytest.MonkeyPatch, discourse: Discourse, b
...
+    mock_get_requests_session.return_value = mocked_session
     mocked_response = mock.MagicMock(spec=requests.Response)
-    mocked_get.return_value = mocked_response
+    mocked_session.get.return_value = mocked_response
     content = "content 1"
     mocked_response.content = content.encode("utf-8")
-    monkeypatch.setattr(requests, "get", mocked_get)
+    monkeypatch.setattr(discourse, "_get_requests_session", mock_get_requests_session)
 
     url = f"{base_path}/t/slug/1"
```
gregory-schiano, line 662: Should use topic url prefix constant

tests/unit/test_discourse.py
```diff
@@ -611,18 +650,43 @@ def test_retrieve_topic(monkeypatch: pytest.MonkeyPatch, discourse: Discourse, b
...
     returned_content = discourse.retrieve_topic(url=url)
 
     assert returned_content == content
 
+    url = "/t/slug/1"
```
gregory-schiano, line 667: Should use topic url prefix constant

tests/unit/test_discourse.py
```diff
@@ -611,18 +650,43 @@ def test_retrieve_topic(monkeypatch: pytest.MonkeyPatch, discourse: Discourse, b
...
+    assert returned_content == content
+
+
+def test_absolute_url(base_path: str, discourse: Discourse):
+    """
+    arrange: given a mocked discourse client
+    act: when absolute_url is called first without the base path and then with it
+    assert: then the url to the topic is returned.
+    """
+    url_path = "/t/slug/1"
```
gregory-schiano, line 679: Should use topic url prefix constant

tests/unit/test___init__.py
```diff
@@ -0,0 +1,113 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Unit tests for execution."""
```
gregory-schiano, line 4: I have the feeling that you're retesting some parts which are already unit tested
I would mock the calls to imported method and only test the "business logic" of the run method given the results of mocked methods

I had the same feeling for integration tests, but since it's integration we need to test all components altogether, but here we're on unit test

discourse_check_cleanup.py
```diff
@@ -0,0 +1,306 @@
...
+        "--check-delete-topics",
+        help="Check that the delete_topics test succeeded",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--check-delete", help="Check that the delete test succeeded", action="store_true"
+    )
+    parser.add_argument(
+        "--check-only", help="Skip cleanup and only run any checks", action="store_true"
+    )
```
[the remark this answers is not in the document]
jdkandersson: Thanks, this is a nice improvement! Should be resolved now

discourse_check_cleanup.py
```diff
@@ -0,0 +1,306 @@
...
+    Args:
+        urls_with_actions: The URLs that had any actions against them.
+        discourse: Client to the documentation server.
+
+    Returns:
+        Whether the test succeeded.
+    """
+    test_name = "create"
+    if not _check_url_count(
+        urls_with_actions=urls_with_actions, expected_count=3, test_name=test_name
```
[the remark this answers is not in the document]
jdkandersson: The trade-off here is complexity. There are 2 kinds of integration tests in this repository, there are those defined in `pytest` and the integration tests defined in the workflow, for which this is a utility. For clarity, let's call these end to end tests since they are intended to simulate how a user would run this action end to end.

The main reasons for the end to end tests to exist are (1) to check that the parameters are correctly being passed from the action input to the code and (2) to check the interactions against the charmhub discourse server rather than a server spun up just for testing. This means that the end to end tests are likely to be fairly stable. Parameters would be useful if these tests were expected to change a lot or we would need to run many different similar tests that just vary by the parameters we add. In this case, we would need to add optional parameters for the input to the tests and then verify whether they have been passed in correctly etc versus simply treating the functions in this utility as the `assert` block for normal tests and the `integration_test.yaml` as the `arrange` and `act` block. It would be great if this utility wasn't needed, although coding these checks in bash would seem to be more complex which is why this utility was created.

Any feedback on this?

src/__init__.py
```diff
@@ -2,3 +2,52 @@
...
+        discourse: A client to the documentation server.
+        dry_run: If enabled, only log the action that would be taken.
+        delete_pages: Whether to delete pages that are no longer needed.
+
+    Returns:
+        All the URLs that had an action with the result of that action.
+
+    """
+    index = get_index(base_path=base_path, server_client=discourse)
+    path_infos = read_docs_directory(docs_path=base_path / "docs")
```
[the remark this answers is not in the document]
jdkandersson: Well spotted, should be fixed shortly

src/__init__.py
```diff
@@ -2,3 +2,52 @@
...
+    server_content = (index.server.content or "") if index.server is not None else ""
```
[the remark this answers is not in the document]
jdkandersson: yes that is cleaner, thanks, should be fixed shortly

src/discourse.py
```diff
@@ -97,7 +99,7 @@ def topic_url_valid(self, url: str) -> _ValidationResult:
             Whether the URL is a valid topic URL.
 
         """
-        if not url.startswith(self._base_path):
+        if not url.startswith(self._base_path) and not url.startswith("/t/"):
```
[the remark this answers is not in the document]
jdkandersson: Good suggestion, thanks, should be fixed shortly

src/discourse.py
```diff
@@ -216,6 +232,18 @@ def _get_post_value(post: dict, key: str, expected_type: type[KeyT]) -> KeyT:
...
+        """Get the URL including base path for a topic.
+
+        Args:
+            url: The relative or absolute URL.
+
+        Returns:
+            The url with the base path.
+        """
+        topic_info = self._url_to_topic_info(url=url)
+        return self._topic_info_to_url(topic_info=topic_info)
```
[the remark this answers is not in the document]
jdkandersson: Yes this could be clearer, the second function name has been changed to `_topic_info_to_absolute_url`. The reason is that the argument to this function could be a relative (starting with `/t/...`) or absolute (starting with https://...) URL. The purpose of this function is that the return value is guaranteed to be the absolute URL, which the conversion to topic info and then subsequent calculation of the URL based on that topic info guarantees. The navigation table stores relative URLs whereas absolute URLs are better for users because they can just click on them from the logs rather than having to figure out the protocol and hostname. Does that make sense? The function name change should be available shortly

src/navigation_table.py
```diff
@@ -9,6 +9,7 @@
 from . import types_
 from .exceptions import NavigationTableParseError
 
+_PUNCTUATION = r"!\"#$%&'()*+,\-./:;<=>?@[\]^_`{|}~"
```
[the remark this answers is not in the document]
jdkandersson: Nice, yes! Should be fixed shortly

tests/integration/test_discourse.py
```diff
@@ -106,27 +101,20 @@ async def test_create_topic_auth_error(
 
 @pytest.mark.asyncio
 async def test_retrieve_topic_auth_error(
-    discourse_user_api_key: str,
     discourse_hostname: str,
     discourse_user_credentials: types.Credentials,
```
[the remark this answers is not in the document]
jdkandersson: Whilst it is possible, what would be the benefits? `discourse_user_credentials` is a pytest fixture that is automatically passed to the test.

The purpose of these series of tests is to check that interactions with discourse fail when expected, which is usually achieved by deliberately using incorrect credentials. See, for example, `test_read_write_permission` which takes an alternate credential. Taking the fixture as a parameter also makes it explicit that the test depends on that fixture rather than implicit vis getting it from `discourse_api`

tests/integration/test___init__.py
```diff
@@ -0,0 +1,273 @@
...
+    assert index_topic == f"{reconcile.NAVIGATION_TABLE_START}".strip()
+    assert index_url in caplog.text
+
+    # 2. docs with an index file in dry run mode
+    caplog.clear()
+    create_metadata_yaml(
+        content=f"{index.METADATA_NAME_KEY}: name 1\n{index.METADATA_DOCS_KEY}: {index_url}",
+        path=tmp_path,
+    )
+    (docs_dir := tmp_path / "docs").mkdir()
```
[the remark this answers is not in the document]
jdkandersson: Well spotted, thanks. Should be fixed shortly

tests/integration/test___init__.py
jdkandersson: Thats a good point, that check should be added shortly in addition to similar checks for the other cases

tests/unit/test_discourse.py
jdkandersson: Thanks, I have updated it here an in a bunch of other places as well, should be available shortly

tests/unit/test___init__.py
jdkandersson: Yes I agree these tests are marginal in the value they provide, I'll write down some thoughts on why they were added and also on mocking.

Starting with mocking, there is value in limiting its use as much as possible because it can get complex quickly and readability goes down the more mocks are used. It also tends to go towards essentially re-writing the source code because the tendency is to check that certain calls did or did not happen and the order etc. The main reasons for mocking are either slow code or calls to external systems in unit tests because this increases the speed of the tests whilst still checking the business logic to a degree. Interactions with external systems should be behind an interface (as they are in this repo for discourse) and have suitable integration tests. After that, the remaining unit tests can reliably mock the external dependency whilst enforcing that the interface spec is being adhered to (which is achieved here using `spec=discourse.Discourse` which raises exceptions if the interface is being used incorrectly, such as calling a non-existent function or forgetting to pass a required variable). The penalty for running other code that is fast or doesn't interact very much with the outside world is relatively minor. If there is a bug with that code, it should cause test failures there which can then be run in isolation using the pytest `-k` option.

The reason to have a small number of unit tests looking at `run` is just to exercise the code a bit and surface any issues around passing the wrong arguments or errors with the logic in that function (e.g., empty versus non empty local index and the dictionary comprehension of reports). If there were no integration tests, it would be necessary to have a lot more unit tests and more extensive interaction checks here. The value of having at least some unit tests is so that obvious mistakes or errors get surfaced more quickly as integration tests take a few minutes to run.

Does that make sense? Is there anything else in particular that would be worth mocking here? Are there any tests that don't provide much value? Are there any others that would be worth adding?

discourse_check_cleanup.py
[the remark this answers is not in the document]
jdkandersson: Fixed

src/__init__.py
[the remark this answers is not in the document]
jdkandersson: Fixed

src/__init__.py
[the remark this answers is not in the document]
jdkandersson: Fixed

src/discourse.py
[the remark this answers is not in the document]
jdkandersson: Fixed

src/discourse.py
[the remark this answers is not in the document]
jdkandersson: Fixed

src/navigation_table.py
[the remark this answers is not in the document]
jdkandersson: Fixed

tests/integration/test___init__.py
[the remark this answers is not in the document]
jdkandersson: Fixed

tests/integration/test___init__.py
jdkandersson: Fixed

tests/unit/test_discourse.py
jdkandersson: Fixed, same for the other comments

tests/unit/test___init__.py
gregory-schiano: The reasoning behind my remark is that scope of unit test should be as limited as possible to avoid test failing in cascade. Let's say a bug is introduced on sub methods, it could result of MANY tests failing in the `run` method while this issue is not lying there at all.
That make finding the root cause more complex, as it could be misleading. That has been part of my conventions for long that unit test should limit their scope to the class they're testing as long as all the external methods are unit tested themselves. We've always looked at which exceptions and returned values are possible, if they influence the caller feedback and mock their behavior in different test scenarios.
This has drawbacks, a change of signature or a new possibly raised exception will require maintenance on these unit tests as they won't test this scenario.
Other convention I've always had also is, we don't test the framework, for example let's say we use an ORM that relies on a mapper to map query results to entities, we don't test that the mapper does its job properly (unless we did some override).

I'm fine with keeping them as they are, if we make sure we stay coherent over the charms over this way of reasoning for unit tests

tests/integration/test_discourse.py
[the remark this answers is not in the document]
gregory-schiano: I'm fine with keeping it like this, I tend to try to minimize the signature size whenever I can, but we're in unit tests and using a fixture so that doesn't really add value

discourse_check_cleanup.py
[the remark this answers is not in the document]
gregory-schiano: I really get the complexity of generalizing these checks. The way I though about it was passing an array as an extra argument or extend the "urls_with_actions" to add the expected result for this url for this test
And then add the logic to count the number of url in the argument and compare the result with the expected result

The shouldn't add too much complexity and allows the test to be more generic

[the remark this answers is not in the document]
jdkandersson: The tests now take an additional parameter `--action-kwargs` which take the expected results for each of the actions which has made them more generic. At this point it does not include a parameter to indicate which URLs should and should not be able to be retrieved which would add a lot more complexity, In all but 1 test it is all URLs should be retrievable anyway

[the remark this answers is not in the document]
jdkandersson: Marking as resolved for now unless there is any further feedback?

src/action.py
```diff
@@ -9,38 +9,57 @@
 from . import exceptions, reconcile, types_
 from .discourse import Discourse
 
-DRAFT_NAVLINK_LINK = "<not created due to draft mode>"
+DRAFT_NAVLINK_LINK = "<not created due to dry run>"
```
arturo-seijas, line 12: Can you order these alphabetically?

src/index.py
```diff
@@ -75,89 +75,55 @@ def _get_key(metadata: dict, key: str) -> str:
     return docs_value
 
 
-def _read_docs_index(base_path: Path) -> str:
+def _read_docs_index(base_path: Path) -> str | None:
```
arturo-seijas, line 78: Wouldn't you rather use `Optional[str]`?

discourse_check_cleanup.py
```diff
@@ -0,0 +1,348 @@
...
+                )
+            )
+        case Action.CLEANUP.value:
+            cleanup(urls_with_actions=urls_with_actions, discourse=discourse, **action_kwargs)
+            sys.exit(0)
+        case _:
+            raise NotImplementedError(f"{args.action} has not been implemented")
+
+
+def _exit_with_result(check_result: bool) -> None:
```
arturo-seijas, line 88: I'd rather throw an exception is an error is detected and the catch it and exit as appropriate instead of controlling the exceptional flow using booleans. You can just raise or re-raise the exception and capture and log at the end of the `match`block.

weiiwang01 (approved):

src/index.py
jdkandersson: Since Python 3.10 syntax for `Union` using the `|` operator has been introduced which has led to developers generally preferring `X | None` instead of `Optional[X]` because (1) it is more explicit in what it means, optional is a bit fuzzy as it could be interpreted as "this variable may not be defined" or "this variable could be `None`" and (2) doesn't require a `typing` import. What do you think? `Optional[X]` is definitely better than `Union[X, None]` for before Python 3.10

discourse_check_cleanup.py
jdkandersson: This utility can be best thought of as a linter. The various `check_...` functions are essentially performing checks similar to linters. A failing check is not necessarily an exception in the flow of the utility as it does not require the check to pass. The `exit_with_code` function is designed to provide an appropriate exit code for GitHub actions to know whether the checks passed or failed, similar to that a tox command could pass or fail. Does that make sense?

discourse_check_cleanup.py
```diff
@@ -0,0 +1,336 @@
...
+import logging
+import sys
+from enum import Enum
+
+from src.discourse import Discourse, create_discourse
+from src.exceptions import DiscourseError
+from src.types_ import ActionResult
+
+
+class Actions(str, Enum):
```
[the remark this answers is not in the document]
jdkandersson: This is a way to make the instances of this class a string. In Python 3.11 there is the `StrEnum` which is equivalent to what is being done here. Once this is upgraded to 3.11, this should be converted to that class. It isn't technically required here, it makes the enums more generally useful so I usually write enums like this be default

src/action.py
jdkandersson: Thanks, fixed

src/index.py
arturo-seijas: Thx for the clarification

discourse_check_cleanup.py
arturo-seijas: It does

arturo-seijas (approved):
LGTM

gregory-schiano (commented):
LGTM

gregory-schiano (approved):
LGTM

discourse_check_cleanup.py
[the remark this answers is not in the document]
jdkandersson: Will resolve this for now, let me know if there is further feedback

Issue #9, opened 2022-11-27
Create a PR for migration of existing docs


jdkandersson:
As an action user where my repo already has documentation on discourse I want the action to download that documentation and create a PR to add it in the `docs` folder so that I can start using the action for documentation in the future.

Acceptance criteria:
- When the action is run on a repository that has the `docs` key in `metadata.yaml` but no `docs` folder, a PR is created on that repository which makes the required changes (e.g., creating files) to be able to start having the documentation in git

jdkandersson:
Closed by #32

PR #22, opened 2022-12-13
Refactor Metadata

yanksyoon:
This PR is a refactor PR in preparation for https://github.com/canonical/upload-charm-docs/issues/9, Migration for existing docs.

Changes:

- change metadata dict to metadata dataclass
- add metadata required fields (name & docs key)
- add metadata required fields testing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -128,6 +128,3 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
-
-# VSCode settings
-.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ coverage.xml
 *.py,cover
 .hypothesis/
 .pytest_cache/
+cov.xml
 
 # Translations
 *.mo
@@ -127,3 +128,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# VSCode settings
+.vscode
--- a/src/metadata.py
+++ b/src/metadata.py
@@ -10,8 +10,8 @@ import yaml
 from . import types_
 from .exceptions import InputError
 
-METADATA_FILENAME = "metadata.yaml"
 METADATA_DOCS_KEY = "docs"
+METADATA_FILENAME = "metadata.yaml"
 METADATA_NAME_KEY = "name"
 
 
--- a/src/metadata.py
+++ b/src/metadata.py
@@ -56,7 +56,7 @@ def get(path: Path) -> types_.Metadata:
         raise InputError(f"Invalid value for name key: {name}, expected a string value")
 
     docs = metadata.get(METADATA_DOCS_KEY)
-    if not isinstance(docs, str | None):
+    if not isinstance(docs, str | None) or (METADATA_DOCS_KEY in metadata and docs is None):
         raise InputError(f"Invalid value for docs key: {docs}, expected a string value")
 
     return types_.Metadata(name=name, docs=docs)
--- a/src/metadata.py
+++ b/src/metadata.py
@@ -1,0 +1,62 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Module for parsing metadata.yaml file."""
+
+from pathlib import Path
+
+import yaml
+
+from . import types_
+from .exceptions import InputError
+
+METADATA_FILENAME = "metadata.yaml"
+METADATA_DOCS_KEY = "docs"
+METADATA_NAME_KEY = "name"
+
+
+def get(path: Path) -> types_.Metadata:
+    """Check for and read the metadata.
+
+    Args:
+        path: The base path to look for the metadata file in.
+
+    Returns:
+        The contents of the metadata file.
+
+    Raises:
+        InputError: if the metadata file does not exists or is malformed.
+
+    """
+    metadata_yaml = path / METADATA_FILENAME
+    if not metadata_yaml.is_file():
+        raise InputError(f"Could not find {METADATA_FILENAME} file, looked in folder: {path}")
+
+    try:
+        metadata = yaml.safe_load(metadata_yaml.read_text())
+    except yaml.error.YAMLError as exc:
+        raise InputError(
+            f"Malformed {METADATA_FILENAME} file, read file: {metadata_yaml}"
+        ) from exc
+
+    if not metadata:
+        raise InputError(f"{METADATA_FILENAME} file is empty, read file: {metadata_yaml}")
+    if not isinstance(metadata, dict):
+        raise InputError(
+            f"{METADATA_FILENAME} file does not contain a mapping at the root, "
+            f"read file: {metadata_yaml}, content: {metadata!r}"
+        )
+
+    if METADATA_NAME_KEY not in metadata:
+        raise InputError(
+            f"Could not find required key: {METADATA_NAME_KEY}, "
+            f"read file: {metadata_yaml}, content: {metadata!r}"
+        )
+    if not isinstance(name := metadata[METADATA_NAME_KEY], str):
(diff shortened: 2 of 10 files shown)


src/metadata.py
```diff
@@ -0,0 +1,62 @@
...
+
+from pathlib import Path
+
+import yaml
+
+from . import types_
+from .exceptions import InputError
+
+METADATA_FILENAME = "metadata.yaml"
+METADATA_DOCS_KEY = "docs"
```
jdkandersson, line 14: Should we sort these alphabetically?

src/types_.py
```diff
@@ -9,6 +9,25 @@
 from pathlib import Path
 from urllib.parse import urlparse
 
+
+class Metadata(typing.NamedTuple):
+    """Information within metadata file. Refer to: https://juju.is/docs/sdk/metadata-yaml.
+
+    Attrs:
+        name: Name of the charm.
+
```
jdkandersson, line 18: Is this intentionally a blank line?

src/types_.py
```diff
@@ -9,6 +9,25 @@
...
+        docs: A link to a documentation cover page on Discourse.
+    """
+
+    # Required fields
```
jdkandersson, line 22: Is this implied by the type? And should this be in the docstring rather than as a comment?

tests/unit/test_index.py
```diff
@@ -153,7 +61,9 @@ def test_get_metadata_yaml_retrieve_discourse_error(tmp_path: Path):
     assert: then ServerError is raised.
     """
     create_metadata_yaml(
-        content=f"{index.METADATA_DOCS_KEY}: http://server/index-page", path=tmp_path
+        content=f"{metadata.METADATA_NAME_KEY}: name\n"
```
jdkandersson, line 64: This indicates a potential refactor of the `index` module since the handling of the `metadata.yaml` file should be in that module, the `index` tests should no longer require creating the file. An option would be passing the metadata in for `index` rather than reading it. What do you think?

tests/unit/test_metadata.py
```diff
@@ -0,0 +1,111 @@
...
+    with pytest.raises(exceptions.InputError) as exc_info:
+        metadata.get(path=tmp_path)
+
+    assert_substrings_in_string(expected_error_msg_contents, str(exc_info.value).lower())
+
+
+@pytest.mark.parametrize(
+    "metadata_name, metadata_docs_url",
+    [
+        pytest.param("name", "", id="name only"),
```
jdkandersson, line 56: What about the case where the `docs` key is not in `metadata.yaml`? These tests seem to check for the case where it is defined (although empty ion one and a URL in the other case).

This could be achieved by restructuring these parameters to be (1) the contents of the `metadata.yaml` file and (2) the expected `Metadata` potentially making use of the walrus operator to avoid having to define constants multiple times. What do you think? It is kind of similar to the next test

Part of the reason is that currently the tests has a bit of logic in it around `docs=metadata_docs_url or None` which is somewhat implicit, in tests its ok to be a bit more verbose for readability reasons

tests/unit/test_index.py
yanksyoon: Yup! I was thinking of pushing those changes with the next PR on migration when the control flow requires moving the handling of metadata.yaml up the flow :) Would that be okay? Otherwise I don't mind pushing those changes with this PR but I thought it would make more sense within the context of migration

src/types_.py
yanksyoon: I wanted to seperate the required/optional fields but shoudl be resolved with the comments below! Thank you!

yanksyoon (commented):
Thanks for the review!

tests/unit/test_metadata.py
yanksyoon: the walrus is awesome! changes applied!

tests/unit/test_metadata.py
```diff
@@ -50,32 +50,46 @@ def test_get_yaml_malformed(
     assert_substrings_in_string(expected_error_msg_contents, str(exc_info.value).lower())
 
 
+# pylint currently does not understand walrus operator perfectly yet
+# pylint: disable=unused-variable,undefined-variable
 @pytest.mark.parametrize(
-    "metadata_name, metadata_docs_url",
+    "metadata_yaml_content, expected_metadata_name, expected_metadata_docs",
```
jdkandersson, line 56: Would this be better if the second parameter was just `expected_metadata`?

tests/unit/test_metadata.py
```diff
@@ -0,0 +1,125 @@
...
+            (
+                "invalid value for",
+                metadata.METADATA_NAME_KEY,
+            ),
+        ),
+        pytest.param(
+            f"{metadata.METADATA_NAME_KEY}: name\n{metadata.METADATA_DOCS_KEY}: 1",
+            ("invalid value for", metadata.METADATA_DOCS_KEY),
+        ),
+    ],
```
jdkandersson, line 110: Should we also check for an empty `docs` key? That likely seems a mistake and will cause an error later anyway when the code tries to retrieve the value

tests/unit/test_index.py
yanksyoon: Yup! change on its way! 👍

.gitignore
```diff
@@ -127,3 +128,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# VSCode settings
+.vscode
```
weiiwang01, line 133: You can set the global git config `core.excludesfile` to exclude IDE files in every project.

yanksyoon: Had no idea that existed! Thank you!

tests/unit/test_metadata.py
yanksyoon: Changed!

tests/unit/test_metadata.py
yanksyoon: Done! Good point!

tests/unit/test_metadata.py
```diff
@@ -0,0 +1,126 @@
...
+            f"{metadata.METADATA_NAME_KEY}: name\n{metadata.METADATA_DOCS_KEY}: ",
+            ("invalid value for", metadata.METADATA_DOCS_KEY),
+        ),
+    ],
+)
+def test_get_invalid_metadata(
+    metadata_yaml_content: str, expected_error_msg_contents: tuple[str, ...], tmp_path: Path
+):
+    """
+    arrange: given metadata with missing required keys
```
gregory-schiano, line 117: I'd add "or invalid value"

tests/unit/test_metadata.py
```diff
@@ -0,0 +1,126 @@
...
+            f"{metadata.METADATA_NAME_KEY}: name\n{metadata.METADATA_DOCS_KEY}: 1",
+            ("invalid value for", metadata.METADATA_DOCS_KEY),
+        ),
+        pytest.param(
+            f"{metadata.METADATA_NAME_KEY}: name\n{metadata.METADATA_DOCS_KEY}: ",
+            ("invalid value for", metadata.METADATA_DOCS_KEY),
+        ),
+    ],
+)
+def test_get_invalid_metadata(
```
gregory-schiano, line 113: I'd put the "non-happy" cases first, and happy case after. So basically moving this method before the `test_get` method

yanksyoon: Done, thanks for the review! :)

tests/unit/test_metadata.py
yanksyoon: Added 👍

arturo-seijas (approved):
LGTM

gregory-schiano (approved):
LGTM

PR #23, opened 2022-12-13
remove type_ attribute and use type instead

### LINKED ISSUE #16: Change action match to be based on type rather than instance variable
The `Action` type currently has a `type_` attribute which is used in `match`, this is duplicate and can be replaced using a match on `type(action)`

jdkandersson:
Changes the match statements to be based on the `type` function rather than the `type_` attribute, closes #16

### DIFF
--- a/src/action.py
+++ b/src/action.py
@@ -274,12 +274,15 @@ def _run_index(
     logging.info("dry run: %s, action: %s", dry_run, action)
 
     if dry_run:
-        if isinstance(action, types_.CreateIndexAction):
-            url = DRY_RUN_NAVLINK_LINK
-        else:
-            url = action.url
         report = types_.ActionReport(
-            table_row=None, url=url, result=types_.ActionResult.SKIP, reason=DRY_RUN_REASON
+            table_row=None,
+            url=(
+                DRY_RUN_NAVLINK_LINK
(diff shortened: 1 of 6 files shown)


yanksyoon (dismissed):
LGTM!

arturo-seijas (approved):
LGTM

src/action.py
```diff
@@ -276,15 +276,19 @@ def _run_index(
...
-            url=DRY_RUN_NAVLINK_LINK if action.type_ == types_.ActionType.CREATE else action.url,
+            url=(
+                DRY_RUN_NAVLINK_LINK
+                if isinstance(action, types_.CreateIndexAction)
+                else action.url
+            ),
             result=types_.ActionResult.SKIP,
             reason=DRY_RUN_REASON,
         )
         logging.info("report: %s", report)
```
yanksyoon, line 287: Unrelated to PR, but may I know when to use "%s" and f-strings?

yanksyoon (approved):
LGTM

src/action.py
arturo-seijas: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#f-strings

merkata (approved):

PR #24, opened 2022-12-13
Fix missing topic bug

### LINKED ISSUE #15: Handle the case where a topic is on the navigation table but it does not exist on the server
Essentially the reconcile for a page locally and on the server fails to retrieve the topic on the server

jdkandersson:
Closes #15

### DIFF
--- a/src/reconcile.py
+++ b/src/reconcile.py
@@ -35,6 +35,32 @@ def _local_only(path_info: types_.PathInfo) -> types_.CreateAction:
     )
 
 
+def _get_server_content(table_row: types_.TableRow, discourse: Discourse) -> str:
+    """Retrieve the content from the server.
+
+    Args:
+        table_row: A row from the navigation table.
+        discourse: A client to the documentation server.
+
+    Returns:
+        The content on the server.
+
+    Raises:
+        ServerError: Retrieving the page contents from the server failed.
+    """
+    if table_row.navlink.link is None:
+        raise exceptions.ReconcilliationError(
+            f"internal error, expecting link on table row, {table_row=!r}"
+        )
+
+    try:
+        return discourse.retrieve_topic(url=table_row.navlink.link).strip()
+    except exceptions.DiscourseError as exc:
+        raise exceptions.ServerError(
+            f"failed to retrieve contents of page, url={table_row.navlink.link}"
+        ) from exc
+
+
 def _local_and_server(
     path_info: types_.PathInfo, table_row: types_.TableRow, discourse: Discourse
 ) -> tuple[
@@ -67,6 +93,7 @@ def _local_and_server(
             row.
         ReconcilliationError: if certain edge cases occur that are not expected, such as
             table_row.navlink.link for a page on the server.
+        ServerError: Retrieving the page contents from the server failed.
     """
     if path_info.level != table_row.level:
         raise exceptions.ReconcilliationError(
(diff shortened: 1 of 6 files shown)


arturo-seijas (dismissed):
LGTM

yanksyoon (dismissed):
LGTM!

src/reconcile.py
```diff
@@ -207,13 +228,18 @@ def _server_only(table_row: types_.TableRow, discourse: Discourse) -> types_.Del
...
+            type_=types_.ActionType.DELETE,
+            level=table_row.level,
+            path=table_row.path,
+            navlink=table_row.navlink,
+            content=discourse.retrieve_topic(url=table_row.navlink.link),
+        )
+    except exceptions.DiscourseError as exc:
+        raise exceptions.ServerError(
+            f"failed to retrieve contents of page, url={table_row.navlink.link}"
+        ) from exc
```
gregory-schiano, line 242: I'd rather enclose only the `retrieve_topic` call in the try/catch block
So something like
```
try:
   content = discourse.retrieve_topic(url=table_row.navlink.link)
except exceptions.DiscourseError as exc:
   ...
return  types_.DeleteAction(...)
```

tests/unit/test_reconcile.py
```diff
@@ -92,6 +92,40 @@ def test__local_and_server_error(
         )
 
 
+def test__get_server_content_missing_link():
+    """
+    arrange: given path info with a file and table row with no changes and discourse client that
```
gregory-schiano, line 97: The arrange description is exactly the same for this test and the next one while their behavior differs in terms of excpected exception
The main difference is the presence of the link in the navlink that changes the behavior of the method. 

I think the arrange description must be adapted

tests/unit/test_reconcile.py
```diff
@@ -92,6 +92,40 @@ def test__local_and_server_error(
...
+        raises an error
+    act: when _get_server_content is called with the path info and table row
+    assert: then ServerError is raised.
```
gregory-schiano, line 100: `ReconcilliationError` and not `ServerError`

src/reconcile.py
jdkandersson: Yes that is better, should be fixed shortly

tests/unit/test_reconcile.py
jdkandersson: Yes, this was not updated as much as needed, should be fixed shortly

tests/unit/test_reconcile.py
jdkandersson: Thanks for spotting this, should be fixed shortly

src/reconcile.py
jdkandersson: Should be fixed now

gregory-schiano (approved):
LGTM

.github/workflows/test.yaml
```diff
@@ -5,7 +5,7 @@ on:
 
 jobs:
   unit-tests:
-    uses: canonical/operator-workflows/.github/workflows/test.yaml@main
+    uses: canonical/operator-workflows/.github/workflows/test.yaml@allow-no-metadata-yaml
```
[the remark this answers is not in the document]
jdkandersson: Yes, I'll hold this PR until then

yanksyoon (dismissed):
LGTM!

.github/workflows/test.yaml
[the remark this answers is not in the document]
jdkandersson: Done

arturo-seijas (approved):
LGTM

gregory-schiano (approved):
LGTM (reapproving)

PR #37, opened 2023-01-16
Fix a bug caused by the checkout working in detached head on GitHub Actions

### LINKED ISSUE #41: Bug: migration fails due to the checkout action using detached head mode
The migration mode will currently fail due to the action resolving the current branch name. This does not work within GitHub actions because the `checkout` action operates in detached head mode. An environment variable should be used instead. Additionally, to avoid this issue from recurring in the future, an e2e test for migration should be introduced.

jdkandersson:
- Resolves #41 
- Adds an e2e test of the migration
- Ensures that only the files in the `docs` folder end up in the migration PR


jdkandersson:
Ready for review now, the failure of the migration e2e test was a discourse internal server error, will retry

arturo-seijas (dismissed):
LGTM

prepare_check_cleanup/migration.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+class Action(str, Enum):
+    """The actions the utility can take.
+
+    Attrs:
+        PREPARE: Prepare discourse pages before running the migration.
+        CHECK_BRANCH: Check that the migration branch was created.
+        CHECK_PULL_REQUEST: Check that the migration pull request was created.
+        CLEANUP: Delete discourse pages before after the migration.
```
gregory-schiano, line 39: `before after` ?
Should probably rephrase :)

prepare_check_cleanup/migration.py
```diff
@@ -0,0 +1,269 @@
...
+        case Action.PREPARE.value:
+            prepare(**action_kwargs)
+            sys.exit(0)
+        case Action.CHECK_BRANCH.value:
+            exit_.with_result(check_branch(**action_kwargs))
+        case Action.CHECK_PULL_REQUEST.value:
+            exit_.with_result(check_pull_request(**action_kwargs))
+        case Action.CLEANUP.value:
+            cleanup(**action_kwargs)
+            sys.exit(0)
```
gregory-schiano, line 82: No validation of kwargs done before passing them to the action methods ?
Would it be interesting ?

prepare_check_cleanup/migration.py
```diff
@@ -0,0 +1,269 @@
...
+            ACTIONS_PULL_REQUEST_TITLE,
+        )
+        return False
+
+    expected_files = {
+        f"{DOCUMENTATION_FOLDER_NAME}/{DOCUMENTATION_INDEX_FILENAME}",
+        f"{DOCUMENTATION_FOLDER_NAME}/page.md",
+    }
+    files_in_pull_request = {file.filename for file in migration_pull_request.get_files()}
+    if expected_files == files_in_pull_request:
```
gregory-schiano, line 219: Shouldn't it be `!=` ?

prepare_check_cleanup/migration.py
```diff
@@ -0,0 +1,269 @@
...
+        logging.exception("cleanup failed for discourse, %s", exc)
+
+    github_repo = _create_repository_client(github_access_token=github_access_token)
+    # Delete the migration PR
+    try:
+        migration_pull_request = _get_migration_pull_request(github_repo=github_repo)
+        if migration_pull_request:
+            migration_pull_request.edit(state="closed")
+    except GithubException as exc:
+        logging.exception("cleanup failed for migration pull request, %s", exc)
```
gregory-schiano, line 258: Hmm I'm wondering if deleting the PR branch doesn't automatically close the PR

tests/unit/test_pull_request.py
```diff
@@ -30,13 +31,37 @@ def test_repository_client__init__(repository: Repo, mock_github_repo: Repositor
...
+        config_reader.get_value(*pull_request.CONFIG_USER_EMAIL) == pull_request.ACTIONS_USER_EMAIL
+    )
+
+
+def test_repository_client__init__name_email_set(repository: Repo, mock_github_repo: Repository):
+    """
+    arrange: given a local git repository client with the user and email configuration already set
+        and mock github repository client
+    act: when RepositoryClient is initialized
+    assert: RepositoryClient is created and git user is not reconfigured.
```
gregory-schiano, line 50: The phrasing `git user is not reconfigured` is kind of strange
Perhaps `git user configuration is kept` ?

tests/unit/test_pull_request.py
```diff
@@ -115,13 +140,21 @@ def test_repository_client_create_branch(
     upstream_repository: Repo,
 ):
     """
-    arrange: given RepositoryClient and newly created files in repo directory
+    arrange: given RepositoryClient and newly created files in repo and repo/docs directory
```
gregory-schiano, line 143: Should be `directories`

tests/unit/test_pull_request.py
```diff
@@ -115,13 +140,21 @@ def test_repository_client_create_branch(
...
     act: when _create_branch is called
-    assert: a new branch is successfully created upstream.
-    """
-    testfile = "testfile.txt"
-    testfile_content = "test"
-    (repository_path / testfile).write_text(testfile_content)
+    assert: a new branch is successfully created upstream with only the files in the repo/docs
```
gregory-schiano, line 145: Shoudn't it be `in the repo and repo/docs directories` ?

prepare_check_cleanup/migration.py
jdkandersson: Yes that is confusing, should be fixed shortly.

prepare_check_cleanup/migration.py
jdkandersson: This is just a CLI to support the e2e testing. If the arguments are badly formed, it will fail which will be obvious in CI. If this was intended for use outside of this repo we should do that for sure. This CLI is also not covered by any tests for the same reason

prepare_check_cleanup/migration.py
jdkandersson: Yes, thanks for spotting this, I think there must still be a bug in the migration e2e tests then

prepare_check_cleanup/migration.py
jdkandersson: Yes it may, it also may complain that there is still an open PR. Either way, i think it is good to be explicit about cleaning up what was created, that way if GitHub changes something we don't suddenly have a whole bunch of PRs lying around

tests/unit/test_pull_request.py
jdkandersson: How about `... git user configuration is not overridden`?

tests/unit/test_pull_request.py
jdkandersson: in this case it is referring to the `repo/docs` directory rather than both in the `repo` and `docs` directory. Does that make sense? I'll add quotes around it to make it more obvious

tests/unit/test_pull_request.py
jdkandersson: Only the files in the `docs` directory within the `repo` should actually be part of the migration branch, if any other files are accidentally created, they should be ignored. This mainly helps for the e2e test which creates some fixtures which ended up in the migration PR before

jdkandersson:
Going back to draft because there seems to be a bug with the migration e2e test not creating some of the expected files

tests/unit/test_pull_request.py
jdkandersson: You are right, I misread which line you were referring to

jdkandersson:
Issues fixed now, ready for another round of reviews

arturo-seijas (approved):
LGTM

tests/unit/test_pull_request.py
gregory-schiano: Good :)

prepare_check_cleanup/migration.py
```diff
@@ -36,7 +36,7 @@ class Action(str, Enum):
         PREPARE: Prepare discourse pages before running the migration.
         CHECK_BRANCH: Check that the migration branch was created.
         CHECK_PULL_REQUEST: Check that the migration pull request was created.
-        CLEANUP: Delete discourse pages before after the migration.
+        CLEANUP: Delete discourse pages and migration pull request and branch after the migration.
```
gregory-schiano, line 39: I would change to `pages, migration pull request and branch` to avoid repeated `and`

weiiwang01 (approved):

prepare_check_cleanup/migration.py
jdkandersson: the second `and` is there to highlight that the `migration` modifier to `pull request` also applies to `branch`, writing it as `pages, migration pull request and branch` can be read that the `migration` modifier only applies to the `pull request` and not `branch`

PR #68, opened 2023-01-20
Operate in temporary directory

### LINKED ISSUE #62: Reset git state after migration action
As an action user, I want the git state after running the migration action to be the original state in which the action was triggered so no unexpected problems occur.

This can be achieved by snapshotting the git state before the action is run and restoring after the action is complete.

jdkandersson:
Closes #62

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -4,121 +4,121 @@ on:
   pull_request:
 
 jobs:
-  # integration-tests:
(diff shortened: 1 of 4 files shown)


arturo-seijas (approved):
LGTM

amandahla (approved):
LGTM

gregory-schiano:
Isn't it possible to do something using git instead ? I'm worried that copying the whole repo could be heavy for bigger repo ?

PR #74, opened 2023-01-23
Url check to include redirect

### LINKED ISSUE #40: `Discourse.topic_url_valid` is too fussy
In `discourse.py` there is a method which checks if Discourse URLs are valid:
https://github.com/canonical/upload-charm-docs/blob/bdaf92b7d6f42b95be3ede01596ec55985a1b4cc/src/discourse.py#L86

A Discourse URL must have the form `/t/topic-slug/<id-num>` to pass validation here, e.g.
https://discourse.charmhub.io/t/get-started-with-the-juju-olm/6559

However, either the slug or topic ID can be omitted, as it can be deduced from the other one. e.g.
https://discourse.charmhub.io/t/6559
https://discourse.charmhub.io/t/get-started-with-the-juju-olm
are both valid URLs, redirecting to the first URL.

This is causing the action to fail when trying to migrate our docs to GitHub. The navigation table contains a link to `/t/6559`, which is failing validation, raising an exception and causing the action to fail.

I would question the need for URL validation at all. When migrating, we should assume that the URLs provided on Discourse are valid, since they already work for existing docs. If a URL really is invalid, then this will be caught later when trying to make a request to it.

jdkandersson:
Closes #40

- Adds an additional check to ensure that the topic URL resolves on discourse
- Relaxes the URL checks to be after any redirects


jdkandersson:
There seems to be a problem, changing this to draft for now

jdkandersson:
Issue fixed, ready for review

tests/integration/test_discourse.py
```diff
@@ -17,6 +17,9 @@
 pytestmark = pytest.mark.discourse
 
 
+SLUG_REGEX = r"\/t\/[\w-]*\/"
```
nrobinaubertin, line 20: why not `r"\/t\/[\w-]+\/"` ?

gregory-schiano (approved):
LGTM

tests/integration/test_discourse.py
jdkandersson: it shouldn't matter in this case although it is probably slightly better. Changed it to using `+`

arturo-seijas (dismissed):
Can we do something about the docker linter suggestion?

merkata (approved):

jdkandersson:
Had to merge changes from main, another review please

gtrkiller (approved):
LGTM

gregory-schiano (approved):
LGTM

jdkandersson:
> Can we do something about the docker linter suggestion?

Thanks for the prompt, I didn't see the comment before merging. This action's use of docker has some opportunities for improvement I have been thinking about. The docker build takes a considerable amount of time now and it would be good to just have an image for the action stored somewhere so it doesn't have the be built every time. This isn't a simple change because it also means that the testing needs to be changed and increases the complexity of releases and the workflow for pull requests. I'll add an issue to improve the use of docker which we can eventually get to.

PR #76, opened 2023-01-23
Fix issue where action required environment variable that is not always available

jdkandersson:
In addition to changelog entry below, moved the self-tests to separate e2e test file because they aren't really integration tests.

### Fixed

- Bug where the action required environment variables that aren't available on all supported triggers.

### Changed

- The check for whether the current branch clashes with the branch being created has been removed.
  This is no longer required because the migration branch is now from the default branch and there
  is an existing check that looks for any clashes with existing branches.
- The action no longer goes back to detached head mode after completing git operations since the
  action now runs in a temporary directory meaning that no changes that persist beyond the action
  completing.


arturo-seijas (approved):
LGTM

.github/workflows/e2e_test.yaml
```diff
@@ -0,0 +1,188 @@
...
+    steps:
+      # Each job has to have this configuration because secrets can be passed through the output of
+      # another job
+      - name: Generate discourse configuration
+        id: configuration
+        run: |
+          echo "discourse= \
+            { \
+              \"hostname\":\"discourse.charmhub.io\", \
+              \"category_id\":\"41\", \
```
merkata, line 18: What is category id `41`, does it make sense to put it in a variable that describes it?

merkata (approved):

.github/workflows/e2e_test.yaml
jdkandersson: Created an issue to track this: https://github.com/canonical/upload-charm-docs/issues/85

PR #97, opened 2023-03-16
fix: integration test

yanksyoon:
Please do not merge, this PR is still in progress.

This PR serves to fix the failing integration tests, to match the upgraded discourse-k8s operator.

### DIFF
--- a/src/discourse.py
+++ b/src/discourse.py
@@ -348,7 +348,7 @@ class Discourse:
 
     @staticmethod
     def _parse_raw_content(content: str) -> str:
-        """Parses raw topic content returned from discourse /raw/{topic_id} API endpoint.
+        """Parse raw topic content returned from discourse /raw/{topic_id} API endpoint.
 
         Args:
             content: Raw content returned by discourse API.
--- a/src/discourse.py
+++ b/src/discourse.py
@@ -379,7 +379,10 @@ class Discourse:
         ) as exc:
             raise DiscourseError(f"Error retrieving the topic, {url=!r}") from exc
 
-        return response.content.decode("utf-8")
+        posts = response.content.decode("utf-8").split("\n\n-------------------------\n\n")
+        if len(posts) < 2:
+            raise DiscourseError(f"No post found in topic, {url=!r} {posts=!r}")
+        return "".join(posts[0].splitlines(keepends=True)[2:])
 
     def create_topic(self, title: str, content: str) -> str:
         """Create a new topic.
--- a/src/discourse.py
+++ b/src/discourse.py
@@ -346,6 +346,25 @@ class Discourse:
         session.mount("https://", adapter)
         return session
 
+    @staticmethod
+    def _parse_raw_content(content: str) -> str:
+        """Parses raw topic content returned from discourse /raw/{topic_id} API endpoint.
+
+        Args:
+            content: Raw content returned by discourse API.
+
+        Returns:
+            Original topic content.
+        """
+        # Discourse version 2.6.0, the content of a topc is returned as raw string.
+        if not content.endswith(_POST_SPLIT_LINE):
+            return content
+
+        # Discourse version 2.8.14, the posts are split by _POST_SPLIT_LINE.
+        posts = content.split(_POST_SPLIT_LINE)
+        post_metadata_removed = posts[0].splitlines(keepends=True)[2:]
+        return "".join(post_metadata_removed)
+
     def retrieve_topic(self, url: str) -> str:
         """Retrieve the topic content.
 
@@ -381,15 +400,7 @@ class Discourse:
             raise DiscourseError(f"Error retrieving the topic, {url=!r}") from exc
 
         content = response.content.decode("utf-8")
-
-        # Discourse version 2.6.0, the content of a topc is returned as raw string.
-        if not content.endswith(_POST_SPLIT_LINE):
-            return content
-
-        # Discourse version 2.8.14, the posts are split by _POST_SPLIT_LINE.
-        posts = content.split(_POST_SPLIT_LINE)
-        post_metadata_removed = posts[0].splitlines(keepends=True)[2:]
-        return "".join(post_metadata_removed)
+        return self._parse_raw_content(content)
 
     def create_topic(self, title: str, content: str) -> str:
         """Create a new topic.
--- a/src/discourse.py
+++ b/src/discourse.py
@@ -401,7 +401,6 @@ class Discourse:
                 category_id=self._category_id,
                 tags=self._tags,
                 content=content,
-                unlist_topic=True,
             )
         except pydiscourse.exceptions.DiscourseError as discourse_error:
             raise DiscourseError(
--- a/src/discourse.py
+++ b/src/discourse.py
@@ -15,6 +15,7 @@ from urllib3 import Retry
 from .exceptions import DiscourseError, InputError
 
 _URL_PATH_PREFIX = "/t/"
+_POST_SPLIT_LINE = "\n\n-------------------------\n\n"
 
 
 class _DiscourseTopicInfo(typing.NamedTuple):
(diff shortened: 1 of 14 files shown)


tests/integration/conftest.py
```diff
@@ -246,29 +302,32 @@ async def discourse_category_id(discourse_client: pydiscourse.DiscourseClient):
...
     discourse_user_credentials: types.Credentials,
-    discourse_hostname: str,
+    discourse_address: str,
     discourse_user_api_key: str,
     discourse_category_id: int,
 ):
     """Create discourse instance."""
     return Discourse(
-        base_path=f"http://{discourse_hostname}",
+        base_path=f"{discourse_address}",
```
jdkandersson, line 311: Is the f-string needed here?

tox.ini
```diff
@@ -97,7 +97,7 @@ commands =
 description = Run integration tests
 deps =
     pytest
-    juju
+    juju==2.9.42.1
```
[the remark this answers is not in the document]
yanksyoon: Yes it does indeed, this also can belong in dev-requirements.txt. Thank you for the feedback!

tests/integration/conftest.py
```diff
@@ -6,233 +6,301 @@
...
+            headers={
+                "Content-Type": "application/x-www-form-urlencoded; charset=UTF-8",
+                "X-CSRF-Token": csrf,
+                "X-Requested-With": "XMLHttpRequest",
+            },
+            data={
+                "login": admin_credentials.email,
+                "password": admin_credentials.password,
+                "second_factor_method": "1",
+                "timezone": "Asia/Hong_Kong",
```
jdkandersson, line 143: Should we make this generic like UTC?

tests/integration/test_discourse.py
```diff
@@ -127,17 +128,17 @@ async def test_retrieve_wrong_url(discourse_api: Discourse):
     url_incorrect_slug = change_url_slug(url, "wrong-slug")
     returned_content = discourse_api.retrieve_topic(url=url_incorrect_slug)
 
-    assert returned_content == content
+    assert content == returned_content
```
jdkandersson, line 131: Same comment as above

tests/integration/test_discourse.py
```diff
@@ -127,17 +128,17 @@ async def test_retrieve_wrong_url(discourse_api: Discourse):
...
 
     url_missing_slug = remove_url_slug(url)
     returned_content = discourse_api.retrieve_topic(url=url_missing_slug)
 
-    assert returned_content == content
+    assert content == returned_content
```
jdkandersson, line 136: same comment as above

tests/integration/test_discourse.py
```diff
@@ -127,17 +128,17 @@ async def test_retrieve_wrong_url(discourse_api: Discourse):
...
 
     url_missing_topic_id = remove_url_topic_id(url)
     returned_content = discourse_api.retrieve_topic(url=url_missing_topic_id)
 
-    assert returned_content == content
+    assert content == returned_content
```
jdkandersson, line 141: same comment as above, also applies to the rest of the switches

jdkandersson:
Thanks for creating this!

src/discourse.py
```diff
@@ -401,7 +411,6 @@ def create_topic(self, title: str, content: str) -> str:
                 category_id=self._category_id,
                 tags=self._tags,
                 content=content,
-                unlist_topic=True,
```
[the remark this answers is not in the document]
yanksyoon: To be exact, only the credentials with appropriate permission level will allow unlisting topics. For users without the permissions, even providing `unlist_topic=False` would make the request fail. :)

tests/integration/conftest.py
```diff
@@ -6,233 +6,301 @@
...
+    Args:
+        discourse_address: The discourse web address.
+        admin_credentials: The discourse admin user credentials.
+
+    Returns:
+        The admin API credentials.
+    """
+    with requests.session() as sess:
+        # Get CSRF token
+        res = sess.get(f"{discourse_address}/session/csrf", headers={"Accept": "application/json"})
```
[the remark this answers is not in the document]
jdkandersson: Isn't there a linting tool that checks for that? Would be great, not that hard to write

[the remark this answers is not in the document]
weiiwang01: I think it's already in our lint checks. I am not sure which one and I am not sure why it's disabled here.

tests/integration/conftest.py
```diff
@@ -6,233 +6,301 @@
...
+        A newly created discourse user credential.
+    """
+    password = secrets.token_urlsafe(16)
+    # Register user
+    requests.post(
+        f"{discourse_address}/users.json",
+        headers={
+            "Api-Key": admin_api_credentials.key,
+            "Api-Username": admin_api_credentials.username,
+        },
```
gregory-schiano, line 185: This is repeated several times I think having a fixture for it could be interesting

yanksyoon: The headers currently and in the foreseeable future doesn't require additional parameters, good idea! Added :)

tests/integration/test_discourse.py
```diff
@@ -75,7 +75,7 @@ async def test_create_retrieve_update_delete_topic(
     url = discourse_api.create_topic(title=title, content=content_1)
     returned_content = discourse_api.retrieve_topic(url=url)
 
-    assert returned_content == content_1, "post was created with the wrong content"
+    assert content_1 == returned_content, "post was created with the wrong content"
```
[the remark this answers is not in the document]
yanksyoon: Reverted the changes! Thank you for the explanation!

src/discourse.py
```diff
@@ -379,7 +380,16 @@ def retrieve_topic(self, url: str) -> str:
         ) as exc:
             raise DiscourseError(f"Error retrieving the topic, {url=!r}") from exc
 
-        return response.content.decode("utf-8")
+        content = response.content.decode("utf-8")
```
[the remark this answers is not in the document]
yanksyoon: Indeed, it is a separate business logic to cater for :) Refactored and added tests, thank you!

tests/integration/conftest.py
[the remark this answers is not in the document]
yanksyoon: I've removed the timeout and ran `tox lint` again, it now does catch the missing timeout
```
tests/integration/conftest.py:215:15: W3101: Missing timeout argument for method 'requests.post' can cause your program to hang indefinitely (missing-timeout)
```

Weird it didn't trigger this time. Thanks for catching that!

weiiwang01 (dismissed):

jdkandersson (dismissed):

jdkandersson:
Looks like integration tests failed due to a fixture name error?

jdkandersson (approved):

tests/unit/test_migration.py
```diff
@@ -602,7 +602,7 @@ def test__migrate_document(tmp_path: Path):
         document_meta=document_meta, discourse=mocked_discourse, docs_path=tmp_path
     )
 
-    assert (file_path := (tmp_path / path)).is_file()
+    assert (file_path := tmp_path / path).is_file()
```
yhaliaw, line 605: Does using the walrus operator here, increase or decrease readability?
I feel the following is easier to read, but does take one more line of code.
```
file_path = tmp_path / path
assert file_path.is_file()
```

yhaliaw (approved):

tests/unit/test_migration.py
yanksyoon: I think it makes the code a tad bit more concise, but I do see your point. Shall we keep the current code as it and refactor it when it becomes complex enough that it would be better to make the `file_path` variable more explicit? :) Thank you for the feedback!

weiiwang01 (approved):

Issue #18, opened 2022-12-11
Expose as command line interface


jdkandersson:
Canonical hosts charm source code in places other than GitHub, make this tool available as a command line interface so that it can be run in those places. Publish on PyPI for easy installation

barrettj12:
Some more advantages:
- The Discourse pattern (navigation tables, etc) used for charm docs is also used by other teams (e.g. Juju). These other teams could use the CLI and write their own actions/workflows to drive it, and we could keep this action specific to charm docs.
- The Discourse -> GitHub docs migration doesn't really make sense as a GitHub action, as it will only be run once. With a CLI this could just be invoked once to pull the Discourse docs into a specified folder, which can be checked into the repo.

jdkandersson:
Won't do this

PR #113, opened 2023-03-22
Community merge conflicts

jdkandersson:
This PR ensures that community contributions are not accidentally overridden by the action.


action.yaml
```diff
@@ -40,6 +40,12 @@ inputs:
       Required if running in migration mode.
     required: false
     type: string
+  base_branch:
```
arturo-seijas, line 43: Perhaps we should default to `main`. Thoughts?

tests/unit/test_repository.py
```diff
@@ -319,6 +319,29 @@ def test_get_file_content(monkeypatch: pytest.MonkeyPatch, repository_client: Cl
     returned_content = repository_client.get_file_content(path=path)
 
     assert returned_content == content
+    mock_github_repository.get_contents.assert_called_once_with(path)
+
+
+def test_get_file_content_branch(monkeypatch: pytest.MonkeyPatch, repository_client: Client):
+    """
+    arrange: given path and branch, Client with a mocked github repository client that returns
```
arturo-seijas, line 327: s/Client/client

tests/unit/test_repository.py
```diff
@@ -319,6 +319,29 @@ def test_get_file_content(monkeypatch: pytest.MonkeyPatch, repository_client: Cl
...
+        content
+    act: when get_file_content is called with the path and branch
+    assert: then the content is returned.
```
arturo-seijas, line 330: s/then the content/the content

tests/unit/test_content.py
```diff
@@ -107,3 +105,32 @@ def test_merge_conflict():
...
+
+@pytest.mark.parametrize(
+    "first, second, expected_diff",
+    _test_diff_parameters(),
+)
+def test_diff(first: str, second: str, expected_diff: str):
+    """
+    arrange: given two strings
+    act: when diff is called with the strings
+    assert: then the expected diff is returned.
```
arturo-seijas, line 132: s/then the/the

arturo-seijas (commented):
I miss a description of the parameters for this action in the readme.

action.yaml
jdkandersson: It is currently optional and results in using the default branch, which could be `main` depending on how the repo is setup. If we set it to `main` and a repo has a different name for the default branch, they the developer needs to manually set it. Does that make sense?

arturo-seijas: Perfect sense. Thx for clarifying

jdkandersson:
Test

src/repository.py
```diff
@@ -0,0 +1,221 @@
...
+            RepositoryClientError: if unexpected error occurred during git operation.
+
+        Returns:
+            True if branch already exists, False otherwise.
+        """
+        try:
+            self._git_repo.git.fetch("origin", branch_name)
+            return True
+        except GitCommandError as exc:
+            if "couldn't find remote ref" in exc.stderr:
```
gregory-schiano, line 82: This forces user to name the remote `origin` couldn't this cause issues ?
That's probably a edge case... what do you think ?

src/repository.py
```diff
@@ -0,0 +1,221 @@
...
+                return False
+            raise RepositoryClientError(
+                f"Unexpected error checking existing branch. {exc=!r}"
+            ) from exc
+
+    def create_branch(self, branch_name: str, commit_msg: str) -> None:
```
gregory-schiano, line 88: This method does more that creating the branch, it creates, commit and push.
I think it could be renamed

src/types_.py
```diff
@@ -37,12 +37,14 @@ class UserInputs(typing.NamedTuple):
             migration mode.
         github_access_token: A Personal Access Token(PAT) or access token with repository access.
             Required in migration mode.
+        base_branch: The branch the the documentation is stored on after updating discourse.
```
gregory-schiano, line 40: `The branch where the documentation is stored` there's a double `the`

gregory-schiano (commented):
First pass, still need to review the tests

src/check.py
```diff
@@ -0,0 +1,101 @@
...
+    """Check whether actions have any content conflicts.
+
+    Args:
+        actions: The actions to check.
+
+    Yields:
+        A problem for each action with a conflict
+    """
+    update_actions = filter(_is_update_action, actions)
+    yield from filter(None, (_update_action_problem(action) for action in update_actions))
```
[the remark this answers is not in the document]
jdkandersson: This is a case of it is generally better to return an iterator than a collection if possible to give the caller a choice on what to do. For example, the caller could chose to stop on the first problem that is encountered or they could choose to return all the problems, if they only want the first or don't want to store all etc, the iterator is less wasteful with resources. Whilst in this case we know the consumer will want all the items, there are few drawbacks with returning an iterator and letting the caller construct the collection as it also puts the code that makes that choice closer to where it makes sense to decide. Does that make sense?

src/content.py
```diff
@@ -0,0 +1,111 @@
...
+        ours: The local change.
+
+    Returns:
+        The description of the merge conflicts or None if there are no conflicts.
+    """
+    try:
+        merge(base=base, theirs=theirs, ours=ours)
+    except ContentError as exc:
+        return str(exc)
+    return None
```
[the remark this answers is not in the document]
jdkandersson: Whilst the logic is relatively simple, the intent of the method is different. the `merge` method is basically intended to be used to merge content that shouldn't have merge conflicts (as used by the `actions` module). This function is intended to let the caller get the conflicts, if any, where a conflict isn't an error. It means that the interface is better for the caller.

If only the `merge` method exists, the caller that just wants to check for conflicts would need to implement exception handling and know that the information about the conflict is in the exception message rather than just being able to ask "what are the conflicts with this" which is a much cleaner interface.

Any thoughts?

src/repository.py
```diff
@@ -0,0 +1,221 @@
...
+    """Wrapper for git/git-server related functionalities."""
+
+    def __init__(self, repository: Repo, github_repository: Repository) -> None:
+        """Construct.
+
+        Args:
+            repository: Client for interacting with local git repository.
+            github_repository: Client for interacting with remote github repository.
+        """
+        self._git_repo = repository
```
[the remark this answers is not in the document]
jdkandersson: We could consider refactoring of this class, it was just moved to a different file in this PR because the class started to be useful not just for creating PRs. I propose that we make this out of scope for this PR

src/repository.py
jdkandersson: I think the way the git repo is initialised is entirely within the control of the action, `origin` is just the default name of the upstream set in git. Any examples you can think of where using `origin` could be problematic?

src/repository.py
jdkandersson: Yeh the name isn't great, we can refactor it in another PR. The history is that it is intended to create the migration branch which is why it is doing the rest as well

src/repository.py
jdkandersson: This is also covered by the "this code was just copied", although if there is a problem it is worth discussing and fixing any bugs ASAP

tests/unit/test_repository.py
jdkandersson: Thanks, should be fixed shortly

tests/unit/test_repository.py
jdkandersson: The `then` is actually used to make the description without the arrange/act/assert read nicely, i.e., given ... when ... then ..., this is actually used quite a lot in the test docstrings

tests/unit/test_reconcile.py
```diff
@@ -167,6 +169,47 @@ def test__local_and_server_file_same(local_content: str, server_content: str, tm
     mock_discourse.retrieve_topic.assert_called_once_with(url=navlink_link)
 
 
+def test__local_and_server_file_content_change_repo_error(tmp_path: Path):
+    """
+    arrange: given path info with a file and table row with no changes and discourse client that
+        returns the different content as in the file and repository that cannot find the file
+    act: when _local_and_server is called with the path info and table row
+    assert: then an update action is returned with None for the base content.
```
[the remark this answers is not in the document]
jdkandersson: as above

tests/unit/test_content.py
jdkandersson: as above

src/action.py
```diff
@@ -140,39 +137,55 @@ def _update(
...
+        if dry_run:
+            result = types_.ActionResult.SKIP
+            reason = DRY_RUN_REASON
+        elif (
+            action.content_change is not None
+            and action.content_change.base is None
+            and action.content_change.new != action.content_change.old
+        ):
+            result = types_.ActionResult.FAIL
+            reason = "no base for the content to be automatically merged"
```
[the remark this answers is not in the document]
jdkandersson: Thanks, should be fixed shortly

src/action.py
```diff
@@ -140,39 +137,55 @@ def _update(
...
     else:
-        result = types_.ActionResult.SKIP if dry_run else types_.ActionResult.SUCCESS
-        reason = DRY_RUN_REASON if dry_run else None
+        if dry_run:
+            result = types_.ActionResult.SKIP
+            reason = DRY_RUN_REASON
+        elif (
+            action.content_change is not None
+            and action.content_change.base is None
+            and action.content_change.new != action.content_change.old
```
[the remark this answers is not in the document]
jdkandersson: Yes correct, the `action` module doesn't know for sure that it won't be called with conflicts so it is better to handle the exception rather than let it fail in a mess

src/check.py
```diff
@@ -0,0 +1,101 @@
...
+        return None
+
+    if action.content_change.base is None:
+        diff = content_diff(action.content_change.old, action.content_change.new)
+        problem = Problem(
+            path=action.path,
+            description=(
+                "cannot execute the update action due to not finding a file on the base branch "
+                "and there are differences between the branch and discourse content, please ensure "
+                f"that there are no differences and try again. Dtected differences:\n{diff}"
```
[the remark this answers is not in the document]
jdkandersson: Thanks, should be fixed shortly

src/check.py
```diff
@@ -0,0 +1,101 @@
...
+            ),
+        )
+    else:
+        action_conflcits = content_conflicts(
```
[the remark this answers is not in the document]
jdkandersson: I don't follow this?

src/content.py
```diff
@@ -0,0 +1,111 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Module for checking for 3-way merge conflicts and create content based on a 3 way merge."""
```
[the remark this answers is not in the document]
jdkandersson: Thanks, should be fixed shortly

src/types_.py
jdkandersson: Thanks, should be fixed shortly

src/check.py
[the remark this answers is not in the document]
gregory-schiano: The variable name has a typo, sorry made a typo in typo 😓

src/check.py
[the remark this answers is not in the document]
gregory-schiano: Well that does make sense in case this is used somewhere else, which is not the case right now. Anyway, this is not a bad implementation, and there's nothing wrong with how that's done, so let's close the comment.

src/repository.py
gregory-schiano: Well indeed, the repo is initialized by the action so that won't cause any problem as long as the default will be "origin", perhaps making it a constant as it's repeated multiple times ?
Or if that's not mandatory to specify the remote (by default it should use the default remote) then just remove "origin" ?

src/repository.py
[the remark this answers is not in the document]
gregory-schiano: Agreed

src/repository.py
gregory-schiano: Ok

src/check.py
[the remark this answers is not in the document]
jdkandersson: thanks, should be fixed shortly

src/content.py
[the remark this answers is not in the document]
gregory-schiano: We it's a bit similar case than the iterator. There's no other usage of the "merge" method that this one, therefore it's more "for future use" and I tend to avoid the "for future use" because the dev that will make the change doesn't necessarily understands the intention behind the choices and could make the change differently.
That more a "philosophy" in such cases I tend to merge the methods and if some refactoring is needed by a future change we'll do it at that time. I apply that philosophy unless I know that a shortly coming change will make use of this method.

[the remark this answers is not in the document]
jdkandersson: There is actually use of `conflicts` in the `check.py` module and `merge` is used in `actions.py`

tests/unit/test_repository.py
```diff
@@ -0,0 +1,441 @@
...
+    mock_github_repo: Repository,
+):
+    """
+    arrange: given valid repository path and a valid access_token and a mocked github client
+    act: when create_repository_client is called
+    assert: RepositoryClient is returned.
+    """
+    origin = git_repo.remote("origin")
+    git_repo.delete_remote(origin)
+    git_repo.create_remote("origin", "https://github.com/test-user/test-repo.git")
```
gregory-schiano, line 430: Why do you need git_repo and make changes here ?

gregory-schiano (commented):
Ok done with the review

src/content.py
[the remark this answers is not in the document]
gregory-schiano: Oh I missed it, sorry the PR is pretty big ^^;

tests/unit/action/test_update_action.py
```diff
@@ -0,0 +1,350 @@
...
+    content: str
+    update_action = src_types.UpdateAction(
+        level=(level := 1),
+        path=(path := "path 1"),
+        navlink_change=src_types.NavlinkChange(
+            old=src_types.Navlink(title="title 1", link=(link := "link 1")),
+            new=src_types.Navlink(title="title 2", link=link),
+        ),
+        content_change=src_types.ContentChange(
+            old=(content := "content 1"), new=content, base=None
```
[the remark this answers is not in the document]
jdkandersson: This particular test is intended to simulate an edge case where the branch has the file and it is on discourse but it isn't on the main branch, in this case only content where both `old` and `new` are the same is allowed through

tests/unit/test_check.py
```diff
@@ -0,0 +1,189 @@
...
+                ExpectedProblem(
+                    path=action_1.path,
+                    description_contents=("conflict", *action_1.content_change[1:]),
+                ),
+                ExpectedProblem(
+                    path=action_2.path,
+                    description_contents=("conflict", *action_2.content_change[1:]),
+                ),
+            ),
+            id="multiple multiple problems",
```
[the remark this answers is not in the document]
jdkandersson: yeh I was debating this when I wrote it, I'll update the others to include `action` as well if they haven't already

tests/unit/test_repository.py
```diff
@@ -0,0 +1,441 @@
...
+    # Check files in the branch
+    branch_files = set(
+        upstream_git_repo.git.ls_tree("-r", branch_name, "--name-only").splitlines()
+    )
+    assert str(root_file) not in branch_files
+    assert str(docs_file) in branch_files
+    assert str(nested_docs_file) in branch_files
+
+
+def test_create_branch_checkout_clash_default(
```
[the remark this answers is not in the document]
jdkandersson: It is because git can confuse the checkout between the file and the branch, if you don't include `--` in the checkout in that case git will actually throw an error. This was actually one of the bugs we found with the migration logic after we first published it, hence the test

tests/unit/action/test_update_action.py
```diff
@@ -0,0 +1,350 @@
...
+    )
+
+    returned_report = action._update(
+        action=update_action, discourse=mocked_discourse, dry_run=True
+    )
+
+    assert_substrings_in_string(
+        (
+            f"action: {update_action}",
+            f"dry run: {True}",
```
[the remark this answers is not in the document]
jdkandersson: I think this used to have a variable for `dry_run` which I'll add back so that the argument on line 93 and this value on line 99 uses the same value

tests/unit/action/test_update_action.py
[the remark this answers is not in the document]
jdkandersson: this is kind of re-using an existing test where only the title changes for this purpose, I'll add a note to the test docstring

tests/unit/action/test_update_action.py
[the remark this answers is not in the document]
jdkandersson: Should be fixed shortly along with the others in this file

tests/unit/test_check.py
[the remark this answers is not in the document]
jdkandersson: Should be fixed with next push

tests/unit/test_repository.py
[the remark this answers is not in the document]
gregory-schiano: Ah yes indeed, got it

src/types_.py
```diff
@@ -252,12 +254,14 @@ class ContentChange(typing.NamedTuple):
...
     Attrs:
+        base: The content which is the base for comparison.
         old: The previous content.
         new: The new content.
     """
 
-    old: Content | None
-    new: Content | None
+    base: Content | None
+    old: Content
```
deusebio, line 263: **thought** I find the naming here a bit confusing. Why is server information the old one (I had to check in the code this to make sure that old was indeed what was coming from discourse)? Really, conceptually the oldest information is probably the base branch (since it does not have either the community contribution or the one proposed by the developer). discourse may even be the most updated version because of a community contribution. My suggestion would be to rename these fields as:

```
base => base
old => server
new => local
```

or

```
base => main
old => discourse
new => proposal
```

src/check.py
```diff
@@ -0,0 +1,102 @@
...
+    Args:
+        action: The action to check.
+
+    Returns:
+        Whether the action is an UpdateAction.
+    """
+    return isinstance(action, UpdateAction)
+
+
+def _update_action_problem(action: UpdateAction) -> Problem | None:
```
deusebio, line 40: **suggestion** can we centralize the "conflicts/check" logic into the ContentChange.conflicts() method (related to another comment here) only? Now it seems that the check is somewhat split between this function and the `content_conflicts`, although really, all the logic also here only considers content_change attribute (and sub-attributes)

For better readability I'd just keep everything in a single place

deusebio (commented):
LGTM. 

I only have some suggestions attached that would (IMHO although I believe this is a somewhat subjective matters) improve readability. 

Rationale: Since conflict checks and merging is conceptually something only involving for `ContentChange`, based on its 3 attribute (`base`, `old`, `new` => although this naming may be a bit confusing, see comment), I believe this may call for centralizing these business logic (check and merge) into class methods

src/content.py
```diff
@@ -0,0 +1,111 @@
...
+        The description of the merge conflicts or None if there are no conflicts.
+    """
+    try:
+        merge(base=base, theirs=theirs, ours=ours)
+    except ContentError as exc:
+        return str(exc)
+    return None
+
+
+def merge(base: str, theirs: str, ours: str) -> str:
```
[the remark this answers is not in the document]
jdkandersson: The reason is that the classes, like `ContentChange`, are more intended to be used to pass data around and define interfaces instead of carrying business logic with them. This repo has generally been functional programming oriented rather than OOP, although in a few cases where it really makes sense OOP is used, typically in cases where a property would just be a transformation of the data which can be a nice thing to add to the data

src/types_.py
jdkandersson: Yeh I was debating this as I was writing the code myself. The reason these attributes are named the way they are is because it previously made sense and I tried to limit the changes as I was starting out. I'm inclined to change the attribute names now anyway since so far a lot of other similar changes have accumulated as well

src/check.py
jdkandersson: The purpose of this method is more than just the conflicts, although that is the use case so far. At a high level, it made sense to introduce this `check` module which takes actions and reports any problems with them. I can see a bunch of other use cases for this, e.g., spellcheck, that links work and so on

src/types_.py
jdkandersson: Should be fixed on the next push, I'll pick server and local to reduce the references to discourse

src/content.py
```diff
@@ -0,0 +1,111 @@
...
+        tmp_path = Path(tmp_dir)
+        repo = Repo.init(tmp_path)
+        writer = repo.config_writer()
+        writer.set_value("user", "name", "temp_user")
+        writer.set_value("user", "email", "temp_email")
+        writer.set_value("commit", "gpgsign", "false")
+        writer.release()
+        base_branch = "base"
+        their_branch = "theirs"
+        our_branch = "ours"
```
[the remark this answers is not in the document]
jdkandersson: Good point, should be fixed on the next push

README.md
```diff
@@ -101,10 +101,10 @@ charmhub.
               github_token: ${{ secrets.GITHUB_TOKEN }}
     ```
 
-    a branch name with `upload-charm-docs/migrate` will be created and a pull 
-    request named `[upload-charm-docs] Migrate charm docs` will be created 
-    towards the working branch the workflow was triggered with. 
-    In order to ensure that the branches can be created successfully, please 
+    a branch name with `upload-charm-docs/migrate` will be created and a pull
+    request named `[upload-charm-docs] Migrate charm docs` will be created
+    towards the working branch the workflow was triggered with.
```
arturo-seijas, line 106: s/towards the working branch/targeting the branch?
It matches better the git terminology

src/__init__.py
```diff
@@ -121,22 +133,14 @@ def run(base_path: Path, discourse: Discourse, user_inputs: UserInputs) -> dict[
     """
     metadata = get_metadata(base_path)
     has_docs_dir = has_docs_directory(base_path=base_path)
+    repository = create_repository_client(
+        access_token=user_inputs.github_access_token, base_path=base_path
+    )
+    clients = Clients(discourse=discourse, repository=repository)
```
arturo-seijas, line 139: What's the reason behind wrapping both clients in the `Client` object?

src/action.py
```diff
@@ -140,39 +138,55 @@ def _update(
...
+    # All of these are needed to ensure types are as expected
+    if (  # pylint: disable=too-many-boolean-expressions
         not dry_run
         and action.navlink_change.new.link is not None
         and action.content_change is not None
-        and action.content_change.new is not None
-        and action.content_change.new != action.content_change.old
+        and action.content_change.base is not None
+        and action.content_change.local is not None
+        and action.content_change.local != action.content_change.server
```
arturo-seijas, line 159: Perhaps I'm missing something but, what if `action.content_change.server` is None?

src/content.py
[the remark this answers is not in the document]
deusebio: Functional and OOP are not really antithetic, in the sense that they can co-exists (e.g. scala, switf, rust). I personally believe it is fine to have domain objects/data classes (that contain data) with helpers methods attached to it. But I understand this is a matter of taste, and although I felt worth suggesting, it is really up to you decide whether you like this, without your feeling that it is ripping apart the style and the structure of the code

src/check.py
```diff
@@ -0,0 +1,102 @@
...
+    """Get any problem with an update action.
+
+    Args:
+        action: The action to check.
+
+    Returns:
+        None if there is no problem or the problem if there is an issue with the action.
+    """
+    if action.content_change is None:
```
arturo-seijas, line 49: This validation logic is the same as the one in actions.py:180

src/check.py
```diff
@@ -0,0 +1,102 @@
...
+        return None
+
+    if (
+        action.content_change.base is None
+        and action.content_change.server == action.content_change.local
+    ):
+        return None
+
+    if action.content_change.base is None:
```
arturo-seijas, line 58: This feels the same as actions.py:181

arturo-seijas (commented):
I've noticed there's a switch case in actions.py:332 executing a different logic based on the Action type. This should instead be implemented with a strategy pattern

src/check.py
deusebio: yes, I see that. I would just keep logic in this as clean and readable as possible, and clearly factor out the different checks based on underlying data/class, e.g. 

```
def _update_action_problem(action: UpdateAction) -> Problem | None:
    action_errors = content_conflicts(action.content_change) or other_check(action.path) or other_check_2(...)
    
    if action_errors is None:
        return None
    
    return Problem(
            path=action.path,
            description=(
                "There are some problems: \n{action_errors}"
            ),
```

Anyhow, as before, this is a bit of matter of personal taste (I believe it increases readability, but other may feel differently), so definitely non-blocking. Resolving discussion

src/constants.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Shared constants."""
```
arturo-seijas, line 4: IMHO these should be defined in docs_directory.py instead. I consider this approach an anti-pattern and usually unjustified as it breaks encapsulation and leads to poor cohesion

src/reconcile.py
```diff
@@ -177,6 +195,14 @@ def _local_and_server(
...
             ),
         )
+
+    try:
+        base_content = clients.repository.get_file_content(
+            path=str(path_info.local_path.relative_to(base_path)),
+            branch=user_inputs.base_branch,
+        )
+    except exceptions.RepositoryClientError:
+        base_content = None
```
arturo-seijas, line 205: Shouldn't we re-raise this exception instead of settings base_content to None? This might be thrown in multiple scenarios, including network issues as per https://github.com/PyGithub/PyGithub/blob/master/github/GithubException.py

src/reconcile.py
```diff
@@ -61,8 +56,36 @@ def _get_server_content(table_row: types_.TableRow, discourse: Discourse) -> str
...
+    table_row: types_.TableRow,
+) -> None:
+    """Input checks before execution.
+
+    Args:
+        path_info: Information about the local documentation file.
+        table_row: A row from the navigation table.
+
+    Raises:
+        ReconcilliationError:
```
arturo-seijas, line 70: Also raises RepositoryClientError

src/__init__.py
jdkandersson: To reduce the number of arguments that need to be passed around, there were many cases of having both discourse and repository as an argument, this simplifies it down to 1 argument

src/action.py
jdkandersson: Actually, you helped identify some unnecessary checks. Both `action.content_change.server` and `action.content_change.local` can no longer be None, it used to be possible but that was changed in this PR

tests/unit/test_repository.py
yanksyoon: This was because the remote on the `git_repo` is local before the changes, which would throw an error when passed to the repository client. The line of code matching the repository is `repository.py:192` where origin url is matched against the regex pattern to get the branch name :)

jdkandersson: ok thanks, I'll add this as a comment

yanksyoon: Thank you!

README.md
jdkandersson: Actually, this is outdated, I'll update it on the next push

src/action.py
jdkandersson: should be fixed in the next push

src/check.py
jdkandersson: It is similar with a different purpose. The validation logic here needs to be run *before* any of the actions are executed. In actions.py, it is a final safeguard before updating the content on discourse

src/constants.py
jdkandersson: I agree, sometimes constants are required in a bunch of places which leads to circular imports that you can only resolve by putting certain constants into a different file. That is the reason for this, the first preference is to keep them with the module. I'll add a note with the reasoning

jdkandersson:
> I've noticed there's a switch case in actions.py:332 executing a different logic based on the Action type. This should instead be implemented with a strategy pattern. Perhaps this is out of scope for this PR, but it'd be good to keep it in mind

There was a similar discussion before. This repo uses these data classes and named tuples primarily for the purpose of sharing data across interfaces rather than to include business logic. Yes it would be possible to implement a function on each of the actions going more towards OOP. The difference in maintainability with that compared to using a match case approach is negligible (I personally would prefer less logic on classes)

src/reconcile.py
jdkandersson: Where do you see that?

src/reconcile.py
jdkandersson: Ah yes, this actually needs more granular exception handling, thanks for pointing that out. Should be fixed on next push.

src/reconcile.py
arturo-seijas: I don't see it anymore with the new changes. Now that exception is always catch and re-raised

dev-requirements.txt
```diff
@@ -1,2 +1,6 @@
 factory_boy>=3.2,<3.3
+ops>=2.1,<2.2
```
arturo-seijas, line 2: Why pinning ops?

src/__init__.py
arturo-seijas: I find it a little obscure, TBH, but I get the point

src/action.py
```diff
@@ -140,39 +138,46 @@ def _update(
...
             result = types_.ActionResult.FAIL
             reason = str(exc)
     else:
-        result = types_.ActionResult.SKIP if dry_run else types_.ActionResult.SUCCESS
-        reason = DRY_RUN_REASON if dry_run else None
+        if dry_run:
+            result = types_.ActionResult.SKIP
+            reason = DRY_RUN_REASON
+        elif (
+            action.content_change is not None
```
arturo-seijas, line 172: It would be good to extract this condition, repeated also in line 151, to a different method for maintainability

src/action.py
```diff
@@ -140,39 +138,46 @@ def _update(
...
             result = types_.ActionResult.SUCCESS
             reason = None
-        except exceptions.DiscourseError as exc:
+        except (exceptions.DiscourseError, exceptions.ContentError) as exc:
             result = types_.ActionResult.FAIL
             reason = str(exc)
     else:
-        result = types_.ActionResult.SKIP if dry_run else types_.ActionResult.SUCCESS
-        reason = DRY_RUN_REASON if dry_run else None
+        if dry_run:
```
arturo-seijas, line 168: Merging this `else` with the nested `if else`block would improve readability. Also, it would probably be simpler if you check first `if dry_run`

src/check.py
arturo-seijas: Considering this validation is executed prior to the the action, does it still make sense to have this scenario in `action._update`?
```
elif (
            action.content_change is not None
            and action.content_change.base is None
            and action.content_change.local != action.content_change.server
        ):
            result = types_.ActionResult.FAIL
            reason = BASE_MISSING_REASON
```
Looks redundant to me

dev-requirements.txt
jdkandersson: Mainly because everything else was pinned, I do agree canonical provided libraries probably shouldn't be pinned, should be removed in next commit

src/action.py
jdkandersson: this condition is actually different, it checks that `action.content_change.base is None` compared to 152 which checks that `action.content_change.base is not None`. Is that what you are referring to?

src/action.py
jdkandersson: yes that is better, should be fixed on next push

src/check.py
jdkandersson: Yes it is still needed since that function accepts any `UpdateAction` and will execute it. That function has no guarantee that it won't get passed an `UpdateAction` that shouldn't actually be updated. There is an edge case that will involve retrieving the content from discourse just before action execution and checking it again so I don't mind having the logic there

src/check.py
jdkandersson: It is similar, although in `action.py` the action is actually getting executed. This is like a pre-run check

src/action.py
jdkandersson: I changed this to make it clearer that action.py executes logic to determine how to execute the action

jdkandersson: Resolving for now, please let me know if you have further feedback on this

src/check.py
jdkandersson: Resolving for now, please let me know if you have further feedback on this

src/check.py
jdkandersson: Resolving for now, please let me know if you have further feedback on this

src/action.py
```diff
@@ -134,45 +180,45 @@ def _update(
...
-    ):
-        raise exceptions.ActionError(f"internal error, new content for page is None, {action=!r}")
 
-    if (
-        action.content_change is not None
-        and action.content_change.old is not None
-        and action.content_change.new is not None
-    ):
-        _log_content_change(old=action.content_change.old, new=action.content_change.new)
+    update_case = _get_update_case(action=action, dry_run=dry_run)
```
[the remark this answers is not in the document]
yanksyoon: Discussed the reason for not centralising validation and the reason for having invalid case (intermediary type conversion may be too verbose and unnecessary).

yanksyoon (commented):
Initial review up to `src/types_.py` done, will start reviewing from `tests/conftests.py` :) Thank you!

.github/workflows/e2e_test.yaml
```diff
@@ -80,6 +83,28 @@ jobs:
...
           echo "docs: ${{ steps.e2e-test-create.outputs.index_url }}" >> metadata.yaml
           cat metadata.yaml
+      - name: Prepare for update test
+        id: e2e-test-prepare-update
+        run: |
+          # Update the contents of one of the files, do it first to get around content conflicts
+          echo -n more content >> docs/doc.md
+          GITHUB_TOKEN='${{ secrets.GITHUB_TOKEN }}'
+          REPO='${{ github.repository }}'
+          PYTHONPATH=$(pwd) python3 prepare_check_cleanup/reconcile.py --action prepare-update --action-kwargs "{\"github_token\": \"$GITHUB_TOKEN\", \"repo\": \"$REPO\", \"filename\": \"docs/doc.md\"}" '${{ steps.e2e-test-create.outputs.urls_with_actions }}' '${{ steps.configuration.outputs.discourse }}'
```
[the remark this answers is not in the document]
jdkandersson: What would be the benefit? Yes this could be improved in a bunch of ways, for example, a more general improvement would be to use something like `click`. Since this is mainly used for e2e test setup and teardown and not exposed to end users it investing in it doesn't seem like it would benefit a lot. At best, it could improve the readability of the e2e test workflow

prepare_check_cleanup/reconcile.py
```diff
@@ -21,13 +28,17 @@ class Action(str, Enum):
     Attrs:
         CHECK_DRAFT: Check that the draft integration test succeeded.
         CHECK_CREATE: Check that the create integration test succeeded.
+        PREPARE_UPDATE: Prepare for the update test.
+        CHECK_UPDATE: Check that the update integration test succeeded.
```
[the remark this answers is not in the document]
jdkandersson: yes, this should be fixed on the next push

prepare_check_cleanup/reconcile.py
```diff
@@ -71,6 +82,15 @@ def main() -> None:
                     urls_with_actions=urls_with_actions, discourse=discourse, **action_kwargs
                 )
             )
+        case Action.PREPARE_UPDATE.value:
+            prepare_update(**action_kwargs)
```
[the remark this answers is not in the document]
jdkandersson: `action_kwargs` is deliberately general in nature since it is intended to provide arguments that are unique to each of the possible actions that can be taken. Type hints are complex and not that useful in those kind of cases, especially since this is mainly used during testing

prepare_check_cleanup/reconcile.py
```diff
@@ -328,17 +426,27 @@ def check_delete(
...
         urls_with_actions: The URLs that had any actions against them.
         discourse: Client to the documentation server.
+        github_token: Token for communication with GitHub.
+        repo: The name of the repository.
     """
     for url in urls_with_actions.keys():
         with contextlib.suppress(DiscourseError):
             discourse.delete_topic(url=url)
 
+    with contextlib.suppress(GithubException):
```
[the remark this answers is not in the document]
jdkandersson: I think this can be changed a bit to have the cleanup log the failure and also report overall success. The suppression is a quick and easy way to ensure all the cleanup actions are taken. Should be implemented in the next push

yanksyoon (approved):
LGTM! Thank you :)

src/repository.py
```diff
@@ -0,0 +1,224 @@
...
+from github import Github
+from github.GithubException import GithubException, UnknownObjectException
+from github.Repository import Repository
+
+from .constants import DOCUMENTATION_FOLDER_NAME
+from .exceptions import InputError, RepositoryClientError, RepositoryFileNotFoundError
+
+GITHUB_HOSTNAME = "github.com"
+ORIGIN_NAME = "origin"
+HTTPS_URL_PATTERN = re.compile(rf"^https?:\/\/.*@?{GITHUB_HOSTNAME}\/(.+\/.+?)(.git)?$")
```
weiiwang01, line 22: What if the remote url is using ssh protocol? For example `git@github.com:canonical/abc.git`.

yanksyoon: By default the github actions uses http protocol, so i think it is safe to parse the HTTPS only

.github/workflows/e2e_test.yaml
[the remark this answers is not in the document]
yanksyoon: I was thinking in terms of reducing maintenance cost by having better readability and structured argument parsing, which gives better visibility within the module on the data that is needed. But after seeing the reasoning for kwargs, i am split between readability issue in json parsing vs having more boilerplate code. Thank you for the reply!

tests/unit/test_repository.py
```diff
@@ -244,39 +245,57 @@ def test_create_pull_request(repository_client: Client, mock_pull_request: PullR
     assert returned_url == mock_pull_request.html_url
 
 
-def test_get_file_content_error(monkeypatch: pytest.MonkeyPatch, repository_client: Client):
+def test_get_file_content_github_error(monkeypatch: pytest.MonkeyPatch, repository_client: Client):
     """
     arrange: given Client with a mocked github repository client that raises an exception
     act: when get_file_content is called
-    assert: RepositoryClientError is raised.
+    assert:  is raised.
```
gregory-schiano, line 252: Probably a typo ?

tests/integration/test___init__run_conflict.py
```diff
@@ -0,0 +1,187 @@
...
+
+from .. import factories
+from ..unit.helpers import assert_substrings_in_string, create_metadata_yaml
+
+pytestmark = pytest.mark.conflict
+
+
+@pytest.mark.asyncio
+@pytest.mark.usefixtures("patch_create_repository_client")
+async def test_run_conflict(
```
gregory-schiano, line 33: Shouldn't we apply the same test but in dry-run ?

src/content.py
```diff
@@ -0,0 +1,112 @@
...
+        writer.set_value("user", "name", "temp_user")
+        writer.set_value("user", "email", "temp_email")
+        writer.set_value("commit", "gpgsign", "false")
+        writer.release()
+
+        # Create base
+        repo.git.checkout("-b", _BASE_BRANCH)
+        (content_path := tmp_path / "content.txt").write_text(base, encoding="utf-8")
+        repo.git.add(".")
+        repo.git.commit("-m", "'initial commit'")
```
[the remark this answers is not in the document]
jdkandersson: You are right, should be fixed on next push

src/content.py
```diff
@@ -0,0 +1,112 @@
...
+        content_path.write_text(ours, encoding="utf-8")
+        repo.git.add(".")
+        repo.git.commit("-m", "'our change'")
+
+        try:
+            repo.git.merge(_THEIR_BRANCH)
+        except GitCommandError as exc:
+            content_conflicts = content_path.read_text(encoding="utf-8")
+            raise ContentError(
+                f"could not automatically merge, conflicts:\n{content_conflicts}"
```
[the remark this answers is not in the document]
jdkandersson: Yes for a future commit, I'll create an issue for this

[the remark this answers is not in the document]
jdkandersson: https://github.com/canonical/upload-charm-docs/issues/131

src/repository.py
jdkandersson: We should consider this for https://github.com/canonical/upload-charm-docs/issues/18, up to now since the action runs in GitHub actions only `https` should be fine for now

src/action.py
```diff
@@ -114,6 +113,56 @@ def _noop(action: types_.NoopAction, discourse: Discourse) -> types_.ActionRepor
...
+    """
+    # Check that action is valid
+    if action.navlink_change.new.link is not None and action.content_change is None:
+        return UpdateCase.INVALID
+    if dry_run:
+        return UpdateCase.DRY_RUN
+    if (
+        not dry_run
+        and action.navlink_change.new.link is not None
+        and action.content_change is not None
```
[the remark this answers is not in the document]
jdkandersson: Yes for defensive coding

tests/unit/test_repository.py
jdkandersson: yes, should be fixed on next push

tests/conftest.py
```diff
@@ -64,17 +64,17 @@ def fixture_repository_path(tmp_path: Path) -> Path:
...
+    upstream_git_repo: Repo,
     upstream_repository_path: Path,
     repository_path: Path,
     test_branch: str,
 ) -> Repo:
     """Create repository with mocked upstream."""
     # uptream_repository is added to create a dependency for the current fixture in order to ensure
     # that the repository can be cloned after the upstream has fully initialized.
-    del upstream_repository
+    del upstream_git_repo
```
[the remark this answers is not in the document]
jdkandersson: I actually don't know what this `del` does, this I think was implemented in a previous PR and this PR just renames it. I'll ask you about it in the stand up and we can change that in a future PR

tests/factories.py
```diff
@@ -7,16 +7,31 @@
...
+
+    # No need for docstring because it is used for type hints
+    def __call__(cls, *args, **kwargs) -> T:  # noqa: N805
+        """Used for type hints of factories."""  # noqa: DCO020
+        return super().__call__(*args, **kwargs)  # noqa: DCO030
+
 
 # The attributes of these classes are generators for the attributes of the meta class
-class PathInfoFactory(factory.Factory):
+# mypy icorrectly believes the factories don't support metaclass
```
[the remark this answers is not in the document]
jdkandersson: thanks, should be fixed on next push

tests/unit/test_reconcile.py
```diff
@@ -201,39 +238,38 @@ def test__local_and_server_file_content_change_repo_error(tmp_path: Path):
...
-    assert returned_action.content_change.new == local_content  # type: ignore
+    assert returned_action.content_change.server == server_content  # type: ignore
+    assert returned_action.content_change.local == local_content  # type: ignore
     assert returned_action.content_change.base is None  # type: ignore
-    mock_discourse.retrieve_topic.assert_called_once_with(url=navlink_link)
-    mock_repository.get_file_content.assert_called_once_with(
-        path=str(path), branch=user_inputs.base_branch
+    mocked_clients.discourse.retrieve_topic.assert_called_once_with(url=navlink_link)
+    mocked_clients.repository.get_file_content.assert_called_once_with(
+        path=filename, branch=user_inputs.base_branch
```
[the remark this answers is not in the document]
jdkandersson: Good point, I'll just call it `relative_path` instead of `filename`. Should be fixed in next push

tests/integration/test___init__run_conflict.py
jdkandersson: Yes good point, I'll add a stage in dry-run mode

tests/conftest.py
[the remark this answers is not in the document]
yanksyoon: I think I might have added this `del` is to remove if from being an unused variable while still using the fixture.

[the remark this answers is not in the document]
jdkandersson: Ah ok, I'll fix that now then

weiiwang01 (approved):

gregory-schiano (approved):
LGTM !
Finally wasn't an easy PR :)

PR #136, opened 2023-04-07
Raise community contribution as PR

deusebio:
Feature described [here](https://docs.google.com/document/d/14LbDoXKug0Jx0wfMEorthMTuR44BApPb38hpqkNanU8/edit#heading=h.jgqqwl2wybuc) and implementation [here](https://docs.google.com/document/d/1Ys-mwkuyNur39IDVVDifJXDfck8G6OpBaarv69PyOZY/edit?usp=sharing)


src/repository.py
```diff
@@ -34,6 +41,47 @@
...
+    is_dirty: bool
+    new: Sequence[str]
+    removed: Sequence[str]
+    modified: Sequence[str]
+
+    @classmethod
+    def from_raw_diff(cls, diffs: Sequence[Diff]):
+        new_files = {diff.a_path for diff in diffs if diff.new_file}
+        removed_files = {diff.a_path for diff in diffs if diff.deleted_file}
+        modified_files = {diff.a_path for diff in diffs if diff.renamed_file}.union(
```
jdkandersson, line 54: Could this be `{diff.a_path for diff in diffs if diff.renamed_file or diff.change_type == "M"}`?

src/repository.py
```diff
@@ -34,6 +41,47 @@
...
+            diff.a_path for diff in diffs if diff.change_type == "M"
+        )
+
+        return DiffSummary(
+            is_dirty=len(diffs) > 0,
+            new=list(new_files),
```
jdkandersson, line 60: Should these be tuples rather than lists since they should be constant?

src/repository.py
```diff
@@ -34,6 +41,47 @@
...
+
+        return DiffSummary(
+            is_dirty=self.is_dirty or other.is_dirty,
+            new=list(set(self.new).union(other.new)),
+            removed=list(set(self.removed).union(other.removed)),
+            modified=list(set(self.modified).union(other.modified)),
+        )
+
+    def __str__(self) -> str:
+        modified_str = [f"modified: {','.join(self.modified)}"] if len(self.modified) > 0 else []
```
jdkandersson, line 77: Should these be tuples as well?

src/repository.py
```diff
@@ -34,6 +41,47 @@
...
+        new_str = [f"new: {','.join(self.new)}"] if len(self.new) > 0 else []
+        removed_str = [f"removed: {','.join(self.removed)}"] if len(self.removed) > 0 else []
+        return " // ".join(
+            modified_str + new_str + removed_str
```
jdkandersson, line 81: f-string?

src/repository.py
```diff
@@ -48,6 +96,92 @@ def __init__(self, repository: Repo, github_repository: Repository) -> None:
...
+    @property
+    def has_docs_directory(self) -> bool:
+        return has_docs_directory(self.base_path)
+
+    @property
+    def current_branch(self) -> str:
+        return self._git_repo.active_branch.name
+
+    @contextmanager
+    def with_branch(self, branch_name: str) -> 'Client':
```
jdkandersson, line 116: Is the quote needed in 3.11?

src/repository.py
```diff
@@ -48,6 +96,92 @@ def __init__(self, repository: Repo, github_repository: Repository) -> None:
...
+        current_branch = self.current_branch
+
+        yield self.switch(branch_name)
```
jdkandersson, line 119: Should this be wrapped in a `try`, `finally`?

src/repository.py
```diff
@@ -48,6 +96,92 @@ def __init__(self, repository: Repo, github_repository: Repository) -> None:
...
+
+        try:
+            self._git_repo.git.checkout(branch_name, "--")
+        finally:
+            if is_dirty:
+                try:
+                    self._git_repo.git.stash("pop")
+                except GitCommandError as exc:
+                    if "CONFLICT" in exc.stdout:
+                        self._git_repo.git.checkout("--theirs", DOCUMENTATION_FOLDER_NAME)
```
jdkandersson, line 152: This nesting is getting a bit deep?

src/repository.py
```diff
@@ -86,39 +220,20 @@ def check_branch_exists(self, branch_name: str) -> bool:
...
-        except GitCommandError as exc:
-            raise RepositoryClientError(f"Unexpected error creating new branch. {exc=!r}") from exc
+    def get_pull_request(self, branch_name: str) -> Optional[str]:
+        open_pull = [
+            pull
+            for pull in self._github_repo.get_pulls(head=branch_name)
+            if pull.head.ref == branch_name
+        ]
+        if len(open_pull) > 1:
+            raise ...
```
jdkandersson, line 230: I'm guessing this isn't done yet?

jdkandersson (commented):
Some initial thoughts

main.py
```diff
@@ -121,21 +141,26 @@ def main() -> None:
...
-        base_path=pathlib.Path(),
-        discourse=discourse,
-        user_inputs=user_inputs,
-    )
+    # Run this only if the reference to discourse exists
+    clients = get_clients(user_inputs, pathlib.Path())
+
+    # Open a PR with community contributions if necessary
+    migrate_result = execute_in_tmpdir(
+        lambda tmp_path: run_migrate(
```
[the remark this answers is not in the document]
deusebio: I don't have strong opinion, but just to double check I understand this correctly (together with the following comment): you mean something like

```python
@execute_in_tmpdir
def main_migrate(repo_path, user_inputs):
    clients = get_clients(user_inputs, repo_path)
    if clients.repository.metadata.docs:
        return run_migrate(clients, user_inputs
    else:
        return {}
```

If so, I need to refactor a bit more `execute_in_tmpdir`, as it needs to change its signature to also deal with functions with other inputs (should this be generic to handle multiple inputs?), e.g. 

```
def execute_in_tmpdir(Callable[[Path, A], B]) -> Callable[A, B]
```

Also, if we want to place the code above in `__init__.py`, we need to move also `execute_in_tmpdir`, otherwise we will end up with circular dependency (but I'm not really sure it is best practice to put so much code in `__init__.py` either),

main.py
```diff
@@ -121,21 +141,26 @@ def main() -> None:
...
+    reconcile_result = execute_in_tmpdir(
+        lambda tmp_path: run_reconcile(
+            clients=get_clients(user_inputs, tmp_path), user_inputs=user_inputs
+        )
+    )() if clients.repository.has_docs_directory else {}
 
     # Write output
-    _write_github_output(urls_with_actions_dict=urls_with_actions_dict)
+    _write_github_output(
+        urls_with_actions_dict=migrate_result | reconcile_result
```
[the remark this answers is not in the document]
deusebio: This should write both outputs, I'm just merging the two results into a single dictionary here. I believe from 3.8 or 3.9, in python the "|" between dictionaries means merge

main.py
```diff
@@ -106,13 +109,30 @@ def wrapper() -> None:
...
+                output = func(execute_cwd)
         finally:
             os.chdir(initial_cwd)
 
+        return output
+
     return wrapper
 
 
+def get_clients(user_inputs: types_.UserInputs, base_path: pathlib.Path) -> Clients:
```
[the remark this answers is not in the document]
deusebio: I don't mind. Where do you think it would be appropriate? `__init__.py`?

src/__init__.py
```diff
@@ -19,25 +19,24 @@
...
 GETTING_STARTED = (
     "To get started with upload-charm-docs, "
     "please refer to https://github.com/canonical/upload-charm-docs#getting-started"
 )
 
 
-def _run_reconcile(
-    base_path: Path, metadata: Metadata, clients: Clients, user_inputs: UserInputs
+def run_reconcile(
+        clients: Clients, user_inputs: UserInputs
```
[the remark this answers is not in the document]
deusebio: yes, I have not yet tried to lint yet. This PR is far from being polished. It was more to share structure, skeleton and main implementation lines

src/__init__.py
```diff
@@ -82,23 +84,15 @@ def _run_reconcile(
...
-        and report.location != DRY_RUN_NAVLINK_LINK
-        and report.location != FAIL_NAVLINK_LINK
+           and report.location != DRY_RUN_NAVLINK_LINK
+           and report.location != FAIL_NAVLINK_LINK
     }
 
 
-def _run_migrate(base_path: Path, metadata: Metadata, clients: Clients) -> dict[str, str]:
-    """Migrate existing docs from charmhub to local repository.
+def download_from_discourse(clients: Clients) -> None:
```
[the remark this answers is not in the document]
deusebio: I'm following your guidance to "where to place what". 

But just one thought that I had during the implementation, was that the `docs_directory.py` seemed rather elementary module, with very few dependency, thus very low in the dependency tree hierarchy. The function `download_from_discourse` function requires quite a number of abstractions (e.g. `Clients` is actually very high in the hierarchy, because it depends on `RepositoryClient` and `Discourse`, cascading to all other modules below), and thus, placing such high-level function in docs_directory would require this module to really move a lot higher in the dependency hierarchical structure.

src/__init__.py
```diff
@@ -112,35 +106,53 @@ def _run_migrate(base_path: Path, metadata: Metadata, clients: Clients) -> dict[
...
-    )
-    clients = Clients(discourse=discourse, repository=repository)
-    if metadata.docs and not has_docs_dir:
-        return _run_migrate(base_path=base_path, metadata=metadata, clients=clients)
-    if has_docs_dir:
-        return _run_reconcile(
-            base_path=base_path, metadata=metadata, clients=clients, user_inputs=user_inputs
+
+    # Remove docs folder and recreate content from discourse
+    clients.repository.switch(user_inputs.base_branch).pull()
```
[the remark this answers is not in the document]
deusebio: Uhm, this does not really have anything to do with docs. We just make sure we are on the `base_branch` and pull any update on this. Why would you put this in the docs_directory module?

src/repository.py
deusebio: Uhm, good point. But since we also don't want items to repeat, maybe frozensets then?

src/repository.py
deusebio: Uhm, I'm not sure how I could get the same result with f-string. I mean here I want to join a list of elements which can be of variable length (size of the list should be between 0 and 3 elements). Do you have a particular form in mind?

src/repository.py
deusebio: Uhm, my IDE was complaining, but maybe it would work anyway...not sure... I'll try this out, but here the problem here is that the function should return the same type of the class this method belongs to. Was this somehow overcome by newer versions of python/mypy?

src/repository.py
deusebio: I don't think so. I mean, this `yield` is to easily define contexts (together with the decorator `contextmanager`), that they can work with 

```
with with_branch(...) as repo:
     # use repo
```

not to catch exception with try-except-finally blocks.

src/repository.py
deusebio: Yes, I agree. I don't like it either. I was meant to refactor this after the first draft

src/repository.py
deusebio: Yes, indeed!

main.py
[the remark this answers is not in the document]
jdkandersson: I think this is fine, I was thinking the 

```python
    if clients.repository.metadata.docs:
        return run_migrate(clients, user_inputs
    else:
        return {}
```

Could go to `run_migrate`, i.e., just get it to return `{}` `if not clients.repository.metadata.docs`. Does that make sense?

src/__init__.py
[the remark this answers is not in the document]
jdkandersson: Sounds good, just wanted to let you know about it 🙂

src/__init__.py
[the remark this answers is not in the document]
jdkandersson: Thats a good thought as well, perhaps it could go into a separate module. The reason I thought `docs_directory.py` would be good is because this function essentially prepares the `docs` directory for further execution. I don't mind moving it up the dependency hierarchy, not sure if it would cause circular dependency problems though in which case a separate module might be better

src/__init__.py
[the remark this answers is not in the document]
jdkandersson: I was thinking that this logic is a little specific for `__init__.py` which should mostly be high level interactions with other modules. The comment is fine, the line of code just seems a little detailed. Perhaps an alternative would be a higher level interface for `repository` that hides some of the detail, e.g., the switch and pull? That sounds a bit like implementation detail

src/repository.py
jdkandersson: yeh sounds good as well

src/repository.py
jdkandersson: oh I misread, I though that `modified_str` was a string. Perhaps the variable name can be clarified, e.g., `modified_files`?

jdkandersson: It is also possible to use `itertools.chain` to avoid needing to create new lists to be able to iterate over all of them, a slight increase in efficiency

src/repository.py
jdkandersson: I can't remember whether they fixed that, I think actually it was supposed to be but then delayed because of some problems in maybe fast API? I'll resolve the comment for now

src/repository.py
jdkandersson: The thought process was that we might want to make sure that the branch is switched back even if an exception occurs during yield so something like

```python
try:
    yield ...
finally:
    # switch branch back
```

So no exception handling, just ensuring that the switch back is executed even if an exception occurs. I haven't done that before, not sure if it works

src/__init__.py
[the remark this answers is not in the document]
jdkandersson: As discussed, we can put this into a module called `download.py`

deusebio:
superseded by https://github.com/canonical/upload-charm-docs/pull/161

PR #137, opened 2023-04-08
Default branch content conflicts

jdkandersson:
Resolves a problem with the original content conflict check where the conflict checks only work in the feature branch and not on the default branch. Instead of the default branch, a specific tag is now used.


action.yaml
```diff
@@ -40,10 +40,16 @@ inputs:
...
+      The name of the tag that contains the documentation after it has been uploaded to
+      discourse.
+    required: false
+    type: string
+    default: "upload-charm-docs/base-content"
   base_branch:
     description: |
-      The name of the branch that contains the documentation after it has been uploaded to
-      discourse. Defaults to the default branch of the repository.
+      DEPRECATED - here for backwards compatibility. This input is no longer used.
```
arturo-seijas, line 52: I think we can drop it given the current adoption

prepare_check_cleanup/reconcile.py
```diff
@@ -192,6 +185,49 @@ def _check_url_result(
...
+
+    Returns:
+        The name of the tag.
+
+    """
+    actions_yaml = Path("action.yaml").read_text(encoding="utf-8")
+    return yaml.safe_load(actions_yaml)["inputs"]["base_tag_name"]["default"]
+
+
+def _check_git_tag_exists(test_name: str, github_repo: Repository, should_exist: bool) -> bool:
```
arturo-seijas, line 199: This method is called always with `should_exist`set to true. Can we remove that argument? Also, it seems a little obscure to me

src/reconcile.py
```diff
@@ -224,6 +229,75 @@ def _local_and_server(
     )
 
 
+def _local_and_server(
```
arturo-seijas, line 232: I don't find this function name very descriptive

src/repository.py
```diff
@@ -141,39 +148,76 @@ def is_dirty(self) -> bool:
         """
         return self._git_repo.is_dirty(untracked_files=True)
 
-    def get_file_content(self, path: str, branch: str | None = None) -> str:
-        """Get the content of a file from the default branch.
+    def tag_commit(self, tag_name: str, commit_sha: str) -> None:
```
arturo-seijas, line 151: If we do this, we'll be tagging twice per commit, one by the upload charm actions and one by this action. I suggest we don't tag as part of the docs upload at all

action.yaml
jdkandersson: The reason I left it in is because our `operator-workflows` use the input for now. If the input is removed in this PR, as soon as it is merged `operator-workflows` will start to fail. Keeping it in for a short time allows us to have `operator-workflows` keep working and be able to switch and then remove the input in a follow up PR. What do you think?

prepare_check_cleanup/reconcile.py
jdkandersson: Thanks for noticing that, yes I had removed the use case for `should_exist=False` and not removed the argument, I'll remove that shortly

src/reconcile.py
jdkandersson: What if it is looked at in context of where it is used in the `_calculate_action` function? An additional note, the name of this function wasn't actually changed in this PR (git is confusing the rename with a new function being added)

prepare_check_cleanup/reconcile.py
jdkandersson: The reason this check exists is because the new way to check for content conflicts is to look at the commit tagged `upload-charm-docs/base-content` (which can be customised) instead of a specific branch. This function checks that the tag has actually been set after the action is run, which is important for the action to work

jdkandersson: Unnecessary argument removed

main.py
```diff
@@ -36,7 +36,10 @@ def _parse_env_vars() -> types_.UserInputs:
     delete_topics = os.getenv("INPUT_DELETE_TOPICS") == "true"
     dry_run = os.getenv("INPUT_DRY_RUN") == "true"
     github_access_token = os.getenv("INPUT_GITHUB_TOKEN")
-    base_branch = os.getenv("INPUT_BASE_BRANCH")
+    # Always defined since the input has a default value
+    base_tag_name = typing.cast(str, os.getenv("INPUT_BASE_TAG_NAME"))
+    # Always defined by GitHub
+    commit_sha = typing.cast(str, os.getenv("GITHUB_SHA"))
```
weiiwang01, line 42: I think you can use `os.environ["INPUT_BASE_TAG_NAME"]` to avoid casting.

jdkandersson: Good point, changed

yanksyoon (approved):
LGTM!

weiiwang01 (approved):

tests/unit/test_repository.py
```diff
@@ -245,107 +250,245 @@ def test_create_pull_request(repository_client: Client, mock_pull_request: PullR
...
+    mock_github_repository.create_git_ref.assert_called_once_with(
+        f"refs/tags/{tag_name}", mock_github_repository.create_git_tag.return_value.sha
+    )
+
+
+def test_tag_commit_tag_exists(monkeypatch: pytest.MonkeyPatch, repository_client: Client):
+    """
+    arrange: given tag name and commit sha, Client with a mocked github repository client
+    act: when tag_commit is called with the tag name and commit sha
+    assert: then the functions are called to create the tag.
```
gregory-schiano, line 335: I'd add `and the function to delete the pre-existing tag is called`

tests/unit/test_repository.py
```diff
@@ -245,107 +250,245 @@ def test_create_pull_request(repository_client: Client, mock_pull_request: PullR
...
+
+    assert_substrings_in_string(("unauthorized", "401"), str(exc.value).lower())
+
+
+def test_get_tag_file_content_unknown_object_error(
+    monkeypatch: pytest.MonkeyPatch, repository_client: Client
+):
+    """
+    arrange: given Client with a mocked github repository client that raises an
+        UnknownObjectException exception during content operations
```
gregory-schiano, line 424: This test and `test_get_tag_file_content_tag_unknown_object_error` are very close each other, can we be more explicit in the "arrange" and specify which method raises the UnknownObjectException for both method ?

src/repository.py
```diff
@@ -141,39 +148,76 @@ def is_dirty(self) -> bool:
...
+
+            # Create the new tag
+            tag_object = self._github_repo.create_git_tag(
+                tag_name, TAG_MESSAGE, commit_sha, "commit"
+            )
+            self._github_repo.create_git_ref(f"refs/tags/{tag_name}", tag_object.sha)
+        except GithubException as exc:
+            raise RepositoryClientError(f"Communication with GitHub failed. {exc=!r}") from exc
+
+    def get_tag_file_content(self, path: str, tag_name: str) -> str:
```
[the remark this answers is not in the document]
jdkandersson: Agreed, should be fixed on next push

src/reconcile.py
```diff
@@ -199,15 +198,21 @@ def _local_and_server(
...
         base_content = None
+    except exceptions.RepositoryTagNotFoundError as exc:
+        raise exceptions.ReconcilliationError(
+            f"Tag {user_inputs.base_tag_name} not defined on the repository, please tag the "
+            "commit with the content matching discourse with the tag "
+            f"{user_inputs.base_tag_name!r}"
+        ) from exc
     except exceptions.RepositoryClientError as exc:
         raise exceptions.ReconcilliationError(
             f"Unable to retrieve content for path from branch, {path=}, "
```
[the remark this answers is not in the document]
jdkandersson: Well spotted, should be fixed on next push

src/repository.py
```diff
@@ -28,6 +34,7 @@
     "existing documentation from server to the git repository."
 )
 PR_LINK_NO_CHANGE = "<not created due to no changes in repository>"
+TAG_MESSAGE = "tag created by upload-charm-docs to mark the latest push to discourse"
```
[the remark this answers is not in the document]
jdkandersson: Good point, changed the wording to indicate that the tag is managed by the action and changes or removal may lead to unexpected behaviour

src/repository.py
```diff
@@ -141,39 +148,76 @@ def is_dirty(self) -> bool:
...
         """
+        # Get the tag
         try:
-            content_file = (
-                self._github_repo.get_contents(path)
-                if branch is None
-                else self._github_repo.get_contents(path, branch)
-            )
+            tag_ref = self._github_repo.get_git_ref(f"tags/{tag_name}")
+            git_tag = self._github_repo.get_git_tag(tag_ref.object.sha)
```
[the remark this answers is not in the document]
jdkandersson: Sounds good, should be fixed on next push

tests/unit/test_repository.py
```diff
@@ -245,107 +250,245 @@ def test_create_pull_request(repository_client: Client, mock_pull_request: PullR
...
-    act: when get_file_content is called
+    arrange: given tag name and commit sha, Client with a mocked github repository client where a
+        deleting a tag raises GithubException
+    act: when tag_commit is called with the tag name and commit sha
+    assert: RepositoryClientError is raised.
+    """
+    mock_github_repository = mock.MagicMock(spec=Repository)
+    mock_github_repository.get_git_ref.return_value.delete.side_effect = GithubException(
+        status=401, data="Unauthorized", headers=None
+    )
```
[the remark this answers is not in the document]
jdkandersson: This can be merged into parametrized, should be fixed on next push

tests/unit/test_repository.py
```diff
@@ -245,107 +250,245 @@ def test_create_pull_request(repository_client: Client, mock_pull_request: PullR
...
 
-def test_get_file_content_unknown_object_error(
+def test_tag_commit_tag_delete_tag_github_error(
     monkeypatch: pytest.MonkeyPatch, repository_client: Client
 ):
     """
-    arrange: given Client with a mocked github repository client that raises an exception
-    act: when get_file_content is called
+    arrange: given tag name and commit sha, Client with a mocked github repository client where a
+        deleting a tag raises GithubException
```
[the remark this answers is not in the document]
jdkandersson: test removed

tests/unit/test_repository.py
jdkandersson: good point, should be fixed in next commit

tests/unit/test_repository.py
jdkandersson: the `arrange` already states `during tag operations` for one and `during content operations` for the other, the reason the method name isn't in the description is that it is possible to see it from the test code and the test description is a bit higher level without including the specific operation name

gregory-schiano: Yes I saw that there was this very small difference, but it’s very difficult to catch the difference at a first sight.
Any idea on how to make the difference more explicit to avoid confusion ?

jdkandersson: the name of the tests have a difference too, I can't think of anything else

gregory-schiano: Ok let’s keep it like this, that’s a small detail

gregory-schiano (approved):
LGTM

weiiwang01 (approved):

PR #156, opened 2023-04-14
handle git commit tags

jdkandersson:
This PR adds support for retrieving content from a tag that is directly attached to a git commit. It would be good to change the `e2e` tests as well, I propose this to be done in a follow up PR as I think using `git` by default creates direct tags rather than indirect tags

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


yhaliaw (approved):

jdkandersson:
Feel free to merge if approved, I'll be away until Tuesday next week

arturo-seijas (approved):
LGTM

gregory-schiano (approved):
LGTM

Issue #174, opened 2023-06-13
Track modifications to `docs` folder only


deusebio:
Currently we allow the tool to potentially commit changes to any of the folder, e.g. see [here](https://github.com/canonical/upload-charm-docs/blob/main/src/repository.py#L255). Given the scope of the action, it is possibly a more sensible option to constrain commits to files in the `docs` folder only

PR #176, opened 2023-06-13
[ISSUE-174] Track modifications to docs folder only

deusebio:
Addressing #174

### DIFF
--- a/src/repository.py
+++ b/src/repository.py
@@ -11,7 +11,7 @@ from contextlib import contextmanager
 from functools import cached_property
 from itertools import chain
 from pathlib import Path
-from typing import Any, NamedTuple, Optional
+from typing import Any, NamedTuple
 
 from git import GitCommandError
 from git.diff import Diff
@@ -216,7 +216,7 @@ class Client:
         finally:
             self.switch(current_branch)
 
-    def get_summary(self, folder: Optional[str] = DOCUMENTATION_FOLDER_NAME) -> DiffSummary:
+    def get_summary(self, folder: str | None = DOCUMENTATION_FOLDER_NAME) -> DiffSummary:
         """Return a summary of the differences against the most recent commit.
 
         Args:
@@ -324,7 +324,7 @@ class Client:
         commit_msg: str,
         push: bool = True,
         force: bool = False,
-        folder: Optional[str] = DOCUMENTATION_FOLDER_NAME,
+        folder: str | None = DOCUMENTATION_FOLDER_NAME,
     ) -> "Client":
         """Update branch with a new commit.
 
--- a/src/repository.py
+++ b/src/repository.py
@@ -216,17 +216,17 @@ class Client:
         finally:
             self.switch(current_branch)
 
-    def get_summary(self, folder: str | None = DOCUMENTATION_FOLDER_NAME) -> DiffSummary:
+    def get_summary(self, directory: str | None = DOCUMENTATION_FOLDER_NAME) -> DiffSummary:
         """Return a summary of the differences against the most recent commit.
 
         Args:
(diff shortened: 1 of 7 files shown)


tests/unit/test_repository.py
```diff
@@ -623,55 +627,96 @@ def test_create_repository_client(
...
     """
     arrange: given repository with a modified file with respect to the head
     act: when get_summary is called
     assert: The DiffSummary object represents the modified file.
     """
-    (repository_client.base_path / "file-1.txt").write_text("My first version")
-    repository_client.update_branch("file-1 commit", force=False, push=False)
+    base_path = repository_client.base_path / folder if folder else repository_client.base_path
+    if folder:
+        base_path.mkdir()
```
gregory-schiano, line 644: I think this can be done through an indirect fixture, which will allow cleanup once the test is done

tests/unit/test_repository.py
```diff
@@ -623,55 +627,96 @@ def test_create_repository_client(
...
     """
     arrange: given repository with a new file with respect to the head
     act: when get_summary is called
     assert: The DiffSummary object represents the new file.
     """
-    (repository_client.base_path / "file-1.txt").write_text("My first version")
-    repository_client.update_branch("file-1 commit", force=False, push=False)
+    base_path = repository_client.base_path / folder if folder else repository_client.base_path
+    if folder:
+        base_path.mkdir()
```
gregory-schiano, line 674: Same comment as above

tests/unit/test_repository.py
```diff
@@ -623,55 +627,96 @@ def test_create_repository_client(
...
     """
     arrange: given repository with a removed file with respect to the head
     act: when get_summary is called
     assert: The DiffSummary object represents the remove file.
     """
-    (repository_client.base_path / "file-1.txt").write_text("My first version")
-    repository_client.update_branch("file-1 commit", force=False, push=False)
+    base_path = repository_client.base_path / folder if folder else repository_client.base_path
+    if folder:
+        base_path.mkdir()
```
gregory-schiano, line 705: Same here

src/repository.py
```diff
@@ -216,13 +216,17 @@ def with_branch(self, branch_name: str) -> Iterator["Client"]:
         finally:
             self.switch(current_branch)
 
-    def get_summary(self) -> DiffSummary:
+    def get_summary(self, folder: str | None = DOCUMENTATION_FOLDER_NAME) -> DiffSummary:
```
jdkandersson, line 219: Nitpick, could we call it `directory`? I think that is the term used elsewhere in this repo

tests/unit/test_repository.py
jdkandersson, line 1: Perhaps I missed it, is there a test that adds a file outside the `docs` directory and the test checks that the file doesn't end up getting pushed to the branch?

tests/unit/test_repository.py
gregory-schiano: Not sure I got your point, I was referencing using the `indirect` parameter to pass the parameter through a fixture
See: https://docs.pytest.org/en/7.1.x/example/parametrize.html#apply-indirect-on-particular-arguments

It's helpful when you have an attribute coming from parametrize that you initialize and want to cleanup after execution of the test

deusebio: Oh, I see. Never used it before. I have added that now. Thanks for the tip

tests/unit/test_repository.py
deusebio: Added!

jdkandersson (approved):

jdkandersson:
Thanks!

gregory-schiano (approved):
LGTM

PR #177, opened 2023-06-15
Update git package to git-svn 1:2.39.2-1.1

arturo-seijas:
The image build is currently failing because the  `git` package is not available anymore

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 FROM python:3.11-slim
 
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends git=1:2.30.* && \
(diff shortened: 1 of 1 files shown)


weiiwang01 (approved):

TeodorPt:
This problem would resurface whenever `python:3.11-slim` is updated again to a new series. Why is the git version pinned? If needed, should the python image also be pinned (e.g. `python:3.11-slim-bookworm`)?

arturo-seijas:
> python:3.11-slim-bookworm

Indeed, we should probably pin the image too

jdkandersson:
actually git is available, it just needs to have a higher version

PR #181, opened 2023-06-23
Close open PR if Discourse and Github content match

deusebio:
I have noticed that a good feature could be to close open PR if github and Discourse matches. This should provide an automatic way to resolve conflicts or also revert PR opened that, given new tags, are no longer necessary

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -131,6 +131,8 @@ def run_migrate(clients: Clients, user_inputs: UserInputs) -> dict[str, str]:
             main_hash = repo.current_commit
         clients.repository.tag_commit(DOCUMENTATION_TAG, main_hash)
 
+    pull_request = clients.repository.get_pull_request(DEFAULT_BRANCH_NAME)
+
     # Check difference with main
     changes = recreate_docs(clients, DOCUMENTATION_TAG)
     if not changes:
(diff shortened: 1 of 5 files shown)


src/__init__.py
```diff
@@ -139,15 +141,18 @@ def run_migrate(clients: Clients, user_inputs: UserInputs) -> dict[str, str]:
             user_inputs.commit_sha,
             DOCUMENTATION_TAG,
         )
+        # If a PR is open we should probably close the PR
```
jdkandersson, line 144: Perhaps `Given there are diffs compared to the base, if a PR is open, it should be closed`?

tests/unit/test___init__.py
```diff
@@ -731,3 +734,59 @@ def test_run_migrate_same_content_local_and_server(caplog, mocked_clients):
 
     assert not returned_migration_reports
     assert any("No community contribution found" in record.message for record in caplog.records)
+    assert not mock_edit_pull_request.called
+
+
+@mock.patch("src.repository.Client.get_pull_request")
```
jdkandersson, line 740: Should we use monkeypatch: https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html

tests/unit/test___init__.py
```diff
@@ -731,3 +734,59 @@ def test_run_migrate_same_content_local_and_server(caplog, mocked_clients):
...
+    caplog.clear()
+    with caplog.at_level(logging.INFO):
+        # run is repeated in unit tests / integration tests
+        returned_migration_reports = run_migrate(
+            clients=mocked_clients, user_inputs=user_inputs
+        )  # pylint: disable=duplicate-code
+
+    assert not returned_migration_reports
+    assert any("No community contribution found" in record.message for record in caplog.records)
+    assert mock_edit_pull_request.called
```
gregory-schiano, line 792: Shouldn't we also test that the method was called with the state attribute equal to "closed" ?

src/__init__.py
deusebio: I'd say

```Given there are NO diffs compared to the base, if a PR is open, it should be closed```

tests/unit/test___init__.py
deusebio: Uhm, I kinda used the one that was used elsewhere in the file to be consistent. It seems that we are using - depending on the file - either monkeypatch or mock.patch, not sure why. But I would keep it consistent per file to start with, but then we should pick one of the two and refactor the others across.

tests/unit/test___init__.py
deusebio: yes, fine.

tests/unit/test___init__.py
jdkandersson: yeh, monkeypatch is preferred. No idea where mock.patch came from

jdkandersson (dismissed):

jdkandersson (dismissed):

jdkandersson (approved):

PR #182, opened 2023-06-29
improve logging of conflicts

jdkandersson:
Improves the logging of conflicts to include new lines

### DIFF
--- a/src/check.py
+++ b/src/check.py
@@ -135,9 +135,13 @@ def _update_action_problem(action: UpdateAction) -> Problem | None:
         )
 
     logging.error(
-        "there is a problem preventing the execution of an action, action: %s, problem: %s",
+        (
+            "there is a problem preventing the execution of an action\npath: %s\nproblem: %s\n"
+            "action: %s"
+        ),
+        problem.path,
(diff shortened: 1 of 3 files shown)


src/check.py
```diff
@@ -135,9 +135,13 @@ def _update_action_problem(action: UpdateAction) -> Problem | None:
         )
 
     logging.error(
-        "there is a problem preventing the execution of an action, action: %s, problem: %s",
+        (
+            "there is a problem preventing the execution of an action\npath: %s\nproblem: %s\n"
+            "action: %s"
```
weiiwang01, line 140: Can we break the long string at the point of newlines? For example
```python
   (
        "there is a problem preventing the execution of an action\n"
        "path: %s\n"
        "problem: %s\n"
        "action: %s"
   )

gregory-schiano (approved):
LGTM

src/check.py
jdkandersson: Good point, fixed

gregory-schiano (approved):
LGTM

weiiwang01 (approved):

PR #183, opened 2023-07-04
[MISC] End-to-End Tests Refactoring

deusebio:
In this PR, we remove the End-2-End tests from the current repository, and moved them to a Test repository, [here](https://github.com/deusebio/repo-test). The reason for this was:

1. Actions are inherently multi-repository, we should have a test repository where we can easily create pre-conditions (e.g. even with tags or branches, without the need to create the files), do commits, merge PRs on which we run actions from a given branch (where the process is implemented). Right now, we should keep on switching branches
2. End-to-end tests interact heavily with the repository (by creating branches, tags, PR), and this would clutter the repository content. Beside, to prevent cluttering, one needs to be able to push force, delete, or other more dev features, that we don't want to enable for a clean, structured and maintanable source repository. Test repositories can have somewhat lighter rules when it comes to the rules of the repositories (e.g. force pushing branches, avoiding signed commits, rolling back). These are all sensible for the base project, but are a bit too restrictive for the test repositories, where we need to play with branches, PRs, etc
3. We should avoid to clutter the main repos with tons of PRs & Triggered Workflows on the main repo as well as Discourse with cluttered content.

The idea discussed with @jdkandersson was to export the E2E tests into a separate repository. Of course this exposes us to the risk of merging features that are not working (altough pass unit and integration tests). The idea here would be to have edge and stable branches in this repository. Scheduled E2E tests workflow run against an edge branch in this repository, whereas actions in production should use stable branches. Promoting edge to stable should be a manual process done once we are comfortable a certain revision is good enough.  

## Tasks

I have now migrated most of the relevant code and business logic about end-to-end testing in the test repository, and I have also extended them to include raising PR and a conflict workflow (only hard conflict so far since the soft ones may go deprecated). This refactoring has however had impacts on this repository as 
1. some small fixes showed up during implementation of E2E tests 
2. we have structured the outputs of the process better (that are used in the E2E tests)
3. Action inputs needed to be extended

There is a [PR](https://github.com/deusebio/repo-test/pull/125/files) which shows some meaningful diff, on how that was achieved, alongside the tests workflow (against the branch of this PR), [here](https://github.com/deusebio/repo-test/actions/runs/5872408897) and [here](https://github.com/deusebio/repo-test/actions/runs/5872408097). The code in the PR can have several improvements, but I believe it would be more critical to get the end-to-end tests up and running.

PR #184, opened 2023-07-06
Sign commits

jdkandersson:
Makes commits signed when required by a given repository.

This requires an e2e tests, they are still being worked on in #183. In the meantime, a successful run is available here: https://github.com/canonical/wordpress-k8s-operator/pull/96

### DIFF
--- a/src/repository.py
+++ b/src/repository.py
@@ -19,6 +19,7 @@ from git.repo import Repo
 from github import Github
 from github.ContentFile import ContentFile
 from github.GithubException import GithubException, UnknownObjectException
+from github.InputGitTreeElement import InputGitTreeElement
 from github.Repository import Repository
 
 from src.docs_directory import has_docs_directory
@@ -132,6 +133,36 @@ class DiffSummary(NamedTuple):
         new_str = (f"new: {','.join(self.new)}",) if len(self.new) > 0 else ()
         removed_str = (f"removed: {','.join(self.removed)}",) if len(self.removed) > 0 else ()
         return " // ".join(chain(modified_str, new_str, removed_str))
+
+
+def _commit_file_to_tree_element(commit_file: commit_module.FileAction) -> InputGitTreeElement:
+    """Convert a file with an action to a tree element.
+
+    Args:
+        commit_file: The file action to convert.
+
+    Returns:
+        The git tree element.
+    """
+    match type(commit_file):
+        case commit_module.FileAdded:
+            commit_file = cast(commit_module.FileAdded, commit_file)
+            return InputGitTreeElement(
+                path=str(commit_file.path), mode="100644", type="blob", content=commit_file.content
+            )
+        case commit_module.FileModified:
+            commit_file = cast(commit_module.FileModified, commit_file)
+            return InputGitTreeElement(
+                path=str(commit_file.path), mode="100644", type="blob", content=commit_file.content
+            )
+        case commit_module.FileDeleted:
+            commit_file = cast(commit_module.FileDeleted, commit_file)
+            return InputGitTreeElement(
+                path=str(commit_file.path), mode="100644", type="blob", sha=None
+            )
+        # Here just in case, should not occur in production
+        case _:  ## pragma: no cover
+            raise NotImplementedError(f"unsupported file in commit, {commit_file}")
 
 
 class Client:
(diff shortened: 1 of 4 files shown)


src/repository.py
```diff
@@ -342,15 +396,37 @@ def update_branch(
...
             Repository client with the updated branch
         """
+        push_args = ["-u"]
+        if force:
+            push_args.append("-f")
+        push_args.extend([ORIGIN_NAME, self.current_branch])
+
         try:
+            # Create the branch if it doesn't exist
+            if push:
```
arturo-seijas, line 406: What's the points of pushing before and after commiting?

src/repository.py
```diff
@@ -342,15 +396,37 @@ def update_branch(
...
-                    args.append("-f")
-                args.extend([ORIGIN_NAME, self.current_branch])
-                self._git_repo.git.push(*args)
+                try:
+                    self._git_repo.git.push(*push_args)
+                except GitCommandError as exc:
+                    # Try with the PyGithub client, suppress any errors and report the original
+                    # problem on failure
+                    try:
+                        logging.info(
```
arturo-seijas, line 418: I think it might be worth to print the exception msg here

src/repository.py
jdkandersson: To create the branch, the PyGithub API doesn't allow for creating branches from specific commits, with a force git push it is allowed. Without pushing the branch first, GitHub actually allows unsigned commits through. By first pushing to create the branch and then pushing the commit, we catch that case and sign the commits in that case too

tests/unit/test_repository.py
```diff
@@ -838,30 +861,146 @@ def side_effect(*args):
...
+
+    repository_client._github_client_push(commit_files=(commit_file,), commit_msg=commit_msg)
+
+    mock_github_repo.get_branch.assert_called_once_with(DEFAULT_BRANCH)
+    mock_github_repo.get_git_tree.assert_called_once_with(
+        sha=mock_github_repo.get_branch.return_value.commit.sha
+    )
+
+    # The InputGitTreeElement does not expose its arguments, can only check that the correct number
+    # were passed
```
gregory-schiano, line 884: Same comment as above

src/repository.py
```diff
@@ -133,6 +135,39 @@ def __str__(self) -> str:
...
+        case commit_module.FileAdded:
+            commit_file = cast(commit_module.FileAdded, commit_file)
+            return InputGitTreeElement(
+                path=str(commit_file.path), mode="100644", type="blob", content=commit_file.content
+            )
+        case commit_module.FileModified:
+            commit_file = cast(commit_module.FileModified, commit_file)
+            return InputGitTreeElement(
+                path=str(commit_file.path), mode="100644", type="blob", content=commit_file.content
+            )
```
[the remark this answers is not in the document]
jdkandersson: yeh I think modified should just be mapped as added, before using the git tree there was a difference but now there isn't a good reason to keep them separate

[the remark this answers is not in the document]
jdkandersson: should be fixed in next push

tests/unit/test_repository.py
```diff
@@ -838,30 +861,146 @@ def side_effect(*args):
...
+
+    (repository_path / DOCUMENTATION_FOLDER_NAME).mkdir()
+    (repository_path / (file_path := Path(f"{DOCUMENTATION_FOLDER_NAME}/test.text"))).write_text(
+        "content 1", encoding="utf-8"
+    )
+
+    mock_git_repository = mock.MagicMock(spec=Git)
+    mock_git_repository.add = mock.Mock(return_value=None)
+    mock_git_repository.commit = mock.Mock(return_value=None)
+    mock_git_repository.show = mock.Mock(return_value=f"A {file_path}")
```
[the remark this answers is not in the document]
jdkandersson: ah yes, thank you! Should be solved in next commit

src/repository.py
arturo-seijas: Thx for the explanation

tests/unit/test_repository.py
```diff
@@ -34,6 +35,28 @@
...
+)
+def test__commit_file_to_tree_element(commit_file: commit.FileAdded):
+    """
+    arrange: given commit file
+    act: when _commit_file_to_tree_element is called with the commit file
+    assert: then a InputGitTreeElement is returned.
+    """
+    tree_element = repository._commit_file_to_tree_element(commit_file=commit_file)
+
+    # InputGitTreeElement don't expose any data, can only check that the correct type is returned
```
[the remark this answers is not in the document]
jdkandersson: thinking about this, it is probably ok to use that, let me try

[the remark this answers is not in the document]
jdkandersson: should be fixed in next push

tests/unit/test_repository.py
jdkandersson: should be fixed on next push

gregory-schiano (approved):
LGTM

yanksyoon (approved):
LGTM!

PR #193, opened 2023-07-19
Contents index hidden

jdkandersson:
The second stage in the [spec](https://discourse.charmhub.io/t/specification-isd003-upload-charm-docs-contents-index-and-navigation-table/7986) adding support for hidden items on the contents index

* Changes the navigation table interpretation to support hidden items (items without a level). The level is inferred by the previous item and assumed to be within a group if the previous item is a group by making the level one more than the group. This does mean that it isn't possible to have a hidden item in an outer group if it immediately follows an empty group. This should not have any impact because hidden items don't show up on the navigation table anyway and are only required to be able to link to them.
* Passes whether the item is hidden via the sorting function where the contents index and the local files are merged
* Adds an integration test for hidden items
* Structural changes to many tests that weren't using factories for test data which was required because the data structures they were using changed

### DIFF
--- a/src/navigation_table.py
+++ b/src/navigation_table.py
@@ -77,7 +77,7 @@ def _line_to_row(line: str) -> types_.TableRow:
     return types_.TableRow(
         level=level,
         path=path,
-        navlink=types_.Navlink(title=navlink_title, link=navlink_link or None),
+        navlink=types_.Navlink(title=navlink_title, link=navlink_link or None, hidden=False),
     )
 
 
--- a/src/navigation_table.py
+++ b/src/navigation_table.py
@@ -153,7 +153,7 @@ def generate_table_row(lines: typing.Sequence[str]) -> typing.Iterator[types_.Ta
         parsed TableRow object, representing the row of the table
     """
     level = 0
-    default_level = 0
+    default_level = 1
     path_components: tuple[str, ...] = ()
 
     for line in lines:
--- a/src/navigation_table.py
+++ b/src/navigation_table.py
@@ -75,6 +75,7 @@ def _line_to_row(line: str, default_level: int) -> types_.TableRow:
     navlink_title = match.group(3)
     navlink_link = match.group(4)
 
+    # Row is marked as hidden if it doesn't have a level
     return types_.TableRow(
         level=level,
         path=path,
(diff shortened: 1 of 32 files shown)


src/navigation_table.py
```diff
@@ -69,15 +70,17 @@ def _line_to_row(line: str) -> types_.TableRow:
...
     path: types_.TablePath = (match.group(2),)
     navlink_title = match.group(3)
     navlink_link = match.group(4)
 
     return types_.TableRow(
         level=level,
         path=path,
-        navlink=types_.Navlink(title=navlink_title, link=navlink_link or None),
+        navlink=types_.Navlink(
+            title=navlink_title, link=navlink_link or None, hidden=match.group(1) is None
```
amandahla, line 82: Do you think is necessary to add a comment making it clear that if a row does not match the pattern will be set as hidden?

jdkandersson: I have added a comment, note that that a row is hidden if it doesn't have a level

yanksyoon (approved):
LGTM!

tests/integration/test___init__run_reconcile.py
```diff
@@ -436,3 +436,230 @@ async def test_run(
...
+    )
+    assert (urls := tuple(urls_with_actions)) == (doc_url, index_url)
+    index_topic = discourse_api.retrieve_topic(url=index_url)
+    assert index_content in index_topic
+    assert doc_table_line_2 in index_topic
+    doc_topic = discourse_api.retrieve_topic(url=doc_url)
+    assert doc_topic == doc_content_3
+
+    # 4. docs with an index file with a hidden documentation file and a new hidden alternate
+    # documentation file
```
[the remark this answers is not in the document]
jdkandersson: it means that there is a second documentation file which is also hidden

tests/factories.py
```diff
@@ -285,6 +251,39 @@ class Params:
     navlink = factory.SubFactory(NavlinkFactory)
 
 
+class ActionReportFactory(
+    factory.Factory, metaclass=BaseMetaFactory[types_.ActionReport]  # type: ignore[misc]
+):
+    """Generate Action reports."""  # noqa: DCO060
```
yhaliaw, line 257: Would it be good to mention DCO060 is from `flake8-docstrings-complete`?

jdkandersson: I think at the top of this file it is extensively documented why these checks are disabled here. I'll see if it can be done at the file level actually

gtrkiller (approved):
LGTM

tests/factories.py
jdkandersson: I'll do this in a follow up PR

PR #194, opened 2023-07-26
Fix navigation table split

jdkandersson:
It seems to be possible to wrap the navigation table using `[details=...]...[/details]` like was done on Wordpress: https://discourse.charmhub.io/t/wordpress-documentation-overview/4052. This means that the navigation table won't be correctly stripped during migration. This PR fixes that problem


yhaliaw (approved):
LGTM

yanksyoon (approved):

PR #202, opened 2023-08-28
[MISC] Update tag for same content

deusebio:
In this PR, I propose a feature to keep updating the base-content tag when reconcile shows that there is no difference between Github and Discourse. 

This feature should allow to enable one way Discourse -> Github sync, even when actions have dry-run flag enabled, to prevent to push content to discourse.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -85,6 +85,21 @@ def run_reconcile(clients: Clients, user_inputs: UserInputs) -> ReconcileOutputs
     # tee creates a copy of the iterator which is needed as check.conflicts consumes the iterator
     # it is passed
     actions, check_actions = tee(actions, 2)
+    if reconcile.is_same_content(index, check_actions):
+        logging.info(
+            "Reconcile not required to run as the content is the same on "
+            "Discourse and Github. Updating the tag."
+        )
+        if clients.repository.is_commit_in_branch(user_inputs.commit_sha, user_inputs.base_branch):
+            # This means we are running from the base_branch
+            clients.repository.tag_commit(DOCUMENTATION_TAG, user_inputs.commit_sha)
+        return ReconcileOutputs(
+            index_url=index.server.url if index.server else "",
+            topics={},
+            documentation_tag=clients.repository.tag_exists(DOCUMENTATION_TAG),
+        )
+
+    actions, check_actions = tee(actions, 2)
     problems = tuple(
         check.conflicts(
             actions=check_actions, repository=clients.repository, user_inputs=user_inputs
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -97,16 +97,16 @@ def run_reconcile(clients: Clients, user_inputs: UserInputs) -> ReconcileOutputs
         return ReconcileOutputs(
             index_url=index.server.url if index.server else "",
             topics=(
-                {clients.discourse.absolute_url(index.server.url): ActionResult.SKIP}
-                if index.server
-                else {}
-            )
-            | (
                 {
                     f"{clients.discourse.absolute_url(row.navlink.link)}": ActionResult.SKIP
                     for row in table_rows
                     if row.navlink.link
                 }
+            )
+            | (
+                {clients.discourse.absolute_url(index.server.url): ActionResult.SKIP}
+                if index.server
+                else {}
             ),
             documentation_tag=clients.repository.tag_exists(DOCUMENTATION_TAG),
         )
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -74,7 +74,7 @@ def run_reconcile(clients: Clients, user_inputs: UserInputs) -> ReconcileOutputs
     sorted_path_infos = sort_module.using_contents_index(
         path_infos=path_infos, index_contents=index_contents, docs_path=docs_path
     )
-    table_rows = navigation_table.from_page(page=server_content, discourse=clients.discourse)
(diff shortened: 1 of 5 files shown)


tests/unit/test_reconcile.py
```diff
@@ -988,3 +988,105 @@ def test_index_page(
...
+                    types_.NavlinkChange(
+                        types_.Navlink("title", None, True), types_.Navlink("title", None, True)
+                    ),
+                    types_.ContentChange(None, "same", "same"),
+                )
+            ],
+            True,
+        ),
+    ],
+)
```
gregory-schiano, line 1031: Could be helpful to have an id for each case to clarify the different cases

tests/unit/test_reconcile.py
```diff
@@ -988,3 +988,105 @@ def test_index_page(
...
+        pytest.param(
+            types_.Index(
+                types_.Page("index-url", "index-content"),
+                types_.IndexFile("index-title", None),
+                "charm",
+            ),
+            False,
+        ),
+    ],
+)
```
gregory-schiano, line 1083: Same comment as previous parametrize

tests/unit/test_reconcile.py
```diff
@@ -988,3 +988,105 @@ def test_index_page(
...
+def test__is_same_content_index(index: types_.Index, expected_value: bool):
+    """
+    arrange: given some mock Index object
```
gregory-schiano, line 1086: add `and an action list that only contains a NoopAction`

tests/unit/test_reconcile.py
```diff
@@ -988,3 +988,105 @@ def test_index_page(
...
+            ],
+            True,
+        ),
+    ],
+)
+def test__is_same_content_actions(
+    actions: typing.Iterable[types_.AnyAction], expected_value: bool
+):
+    """
+    arrange: given some mock list of actions
```
gregory-schiano, line 1036: Add `and an index with matching local and server content`

deusebio:
@gregory-schiano thanks for the review! I believe I have addressed your comments!

src/__init__.py
```diff
@@ -84,6 +84,35 @@ def run_reconcile(clients: Clients, user_inputs: UserInputs) -> ReconcileOutputs
 
     # tee creates a copy of the iterator which is needed as check.conflicts consumes the iterator
     # it is passed
+    actions, check_actions = tee(actions, 2)
+    if reconcile.is_same_content(index, check_actions):
```
jdkandersson, line 88: As discussed in mattermost, I think we just need to change the check on line 113 of `__init__.py` be something like `not user_inputs.dry_run or <all actions noop>` and that should get us what we need - any update action should be considered as a difference event if the content doesn't change because the navigation table would be changing in that case

deusebio: I have reverted the business logic to only execute if there are NoopActions, instead of also considering UpdateAction with no content change, that, as discussed, may mean a different in the tree structure

However I believe it is actually sane to leave the check and the business logic where it is. We would nevertheless need to `tee` the iterator, and also having checks there can prevent the code execution after which is totally useless if there are only NoopAction. Beside this should apply regardless if the action is running in dry mode or not

tests/unit/test___init__.py
```diff
@@ -169,6 +169,118 @@ def test__run_reconcile_local_empty_server(mocked_clients):
...
+        Args:
+            path: path of the file
+            tag_name: name of the tag
+
+        Returns:
+            Content of the given tag
+        """
+        assert tag_name
+
+        if path == str((docs_folder / "index.md").relative_to(repository_path)):
```
jdkandersson, line 230: Should the path for this file be a variable that is re-used here? E.g.,  `(index_path := (docs_folder / "index.md")).write_text(...)` - also applicable for any other similar code

tests/unit/test_reconcile.py
```diff
@@ -1035,3 +1036,116 @@ def test_index_page(
     returned_action = reconcile.index_page(index=index, table_rows=table_rows)
 
     assert returned_action == expected_action
+
+
+@pytest.mark.parametrize(
+    "actions, expected_value",
+    [
+        pytest.param(
+            [types_.NoopAction(1, tuple("path"), types_.Navlink("title", None, True), None)],
```
jdkandersson, line 1045: should the factories be used for these instead?

gregory-schiano (dismissed):
Had another look LGTM
I agree with the comment from David for using the factories

tests/unit/test_reconcile.py
deusebio: Definitely. I did not see the factories for this! Thanks!

tests/unit/test_reconcile.py
```diff
@@ -1041,42 +1041,14 @@ def test_index_page(
...
-        ),
-        pytest.param(
-            [types_.CreateAction(1, tuple("path"), "title", None, True)], False, id="Create action"
-        ),
-        pytest.param(
-            [types_.DeleteAction(1, tuple("path"), types_.Navlink("title", None, True), None)],
-            False,
-            id="Delete action",
-        ),
+        pytest.param([factories.NoopActionFactory()], True, id="Noop action"),
```
jdkandersson, line 1044: So much nicer!

jdkandersson (approved):

tests/unit/test_reconcile.py
deusebio: Indeed!

gregory-schiano (approved):
LGTM

PR #205, opened 2023-09-01
Strip base_content in reconcile

### LINKED ISSUE #201: Conflict arising due to trailing newlines in Discourse posts
It seems that it is not possible to write Discourse posts that contain trailing new lines. This leads to a conflict in the `upload-charm-docs` action in a PR when modifying a document where the base version 
(the version in the commit tagged with `upload-charm-docs/base-content`) is the same as the Discourse post, except that it contains trailing new lines (or probably ws chars in general).

If the document is not modified in a PR, no conflict is detected because trailing ws characters are stripped from a local version: https://github.com/canonical/upload-charm-docs/blob/3cb5764f479a08ec067c621aaeb2f2856cef7fd4/src/reconcile.py#L191

As opposed to the base content: 

https://github.com/canonical/upload-charm-docs/blob/3cb5764f479a08ec067c621aaeb2f2856cef7fd4/src/reconcile.py#L210-L212

My current understanding is that ws characters should also be stripped from the base content (in fact, adding `strip()` to this line also passes all the tests) like the local version.

cbartz:
Remove leading/trailing whitespace from base content in the same way as from local content.

Fixes #201

### DIFF
--- a/src/reconcile.py
+++ b/src/reconcile.py
@@ -209,7 +209,7 @@ def _local_and_server_file_local_page_server(
         path = str(path_info.local_path.relative_to(base_path))
         base_content = clients.repository.get_file_content_from_tag(
             path=path, tag_name=DOCUMENTATION_TAG
-        )
+        ).strip()
     except exceptions.RepositoryFileNotFoundError:
         base_content = None
     except exceptions.RepositoryTagNotFoundError as exc:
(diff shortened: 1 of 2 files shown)


jdkandersson (approved):
Thanks!

arturo-seijas (approved):
LGTM

PR #207, opened 2023-09-05
add PR checklist and generate source docs

jdkandersson:
Applicable spec: N/A

### Overview

Adds the `src-docs` and updates the documentation to indicate that the action should only be used after checking with the docs team

### Rationale

To make it easier to review future PRs

### Module Changes

No changes

### Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,8 @@
+## Generating src docs for every commit
+
+Run the following command:
+
+```bash
+echo -e "tox -e src-docs\ngit add src-docs\n" > .git/hooks/pre-commit
+chmod +x .git/hooks/pre-commit
+```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,6 +3,6 @@
 Run the following command:
 
 ```bash
-echo -e "tox -e src-docs\ngit add src-docs\n" > .git/hooks/pre-commit
+echo -e "tox -e src-docs\ngit add src-docs\n" >> .git/hooks/pre-commit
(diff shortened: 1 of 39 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,8 @@
+## Generating src docs for every commit
+
+Run the following command:
+
+```bash
+echo -e "tox -e src-docs\ngit add src-docs\n" > .git/hooks/pre-commit
```
cbartz, line 6:
```suggestion
echo -e "tox -e src-docs\ngit add src-docs\n" >> .git/hooks/pre-commit
```
Might be a bit nitpicky, but I prefer not overriding any existing hooks.

cbartz (dismissed):
LGTM . Just a small suggested change.

cbartz (approved):

generate-src-docs.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+rm -rf src-docs
+lazydocs --no-watermark --output-path src-docs src/*.py
```
yanksyoon, line 7: Out of curiosity, what is the reason for having *.py please? wouldn't src/* be enough?

yanksyoon (approved):
LGTM!

merkata (approved):
lgtm

generate-src-docs.sh
jdkandersson: it is so that each file is considered separately rather than together - mostly this is for the charm rather than in this case

PR #209, opened 2023-09-05
[MISC] Fixing bug arising when repositories are shallow

deusebio:
The actions are most-likely failing across repositories, since we have introduced a check to make sure that tags do indeed belong to the base branch. 

However, git repositories in Github Actions are shallow, only carrying a single commit. Therefore, the checks may erroneously fail since the documentation tag commit is unknown to the base branch. 

While doing this I have also slightly refactored the `pre_flight_check` to provide some more meaningful logging

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -225,10 +225,26 @@ def pre_flight_checks(clients: Clients, user_inputs: UserInputs) -> bool:
     Returns:
         Boolean representing whether the checks have all been passed.
     """
-    with clients.repository.with_branch(user_inputs.base_branch) as repo:
-        if repo.tag_exists(DOCUMENTATION_TAG):
-            return repo.is_commit_in_branch(
-                repo.switch(DOCUMENTATION_TAG).current_commit, user_inputs.base_branch
+    clients.repository.switch(user_inputs.base_branch)
+
+    if documentation_commit := clients.repository.tag_exists(DOCUMENTATION_TAG):
+        flag = clients.repository.is_commit_in_branch(
+            documentation_commit, user_inputs.base_branch
+        )
+        if not flag:
+            logging.error(
+                "Inconsistent repository: documentation tag %s (at commit %s)"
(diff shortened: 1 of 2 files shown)


src/repository.py
```diff
@@ -280,6 +280,7 @@ def is_commit_in_branch(self, commit_sha: str, branch: str | None = None) -> boo
         """
         star_pattern = re.compile(r"^\* ")
         try:
+            self._git_repo.git.fetch("--depth=2147483647")
```
deusebio, line 283: > The special depth 2147483647 (or 0x7fffffff, the largest positive number a signed 32-bit integer can contain) means infinite depth.

Reference [here](https://git-scm.com/docs/shallow)

jdkandersson: Perhaps make this a comment on the code?

jdkandersson (approved):

src/__init__.py
```diff
@@ -225,10 +225,26 @@ def pre_flight_checks(clients: Clients, user_inputs: UserInputs) -> bool:
...
-        return True
+        return flag
+
+    logging.info(
+        "documentation tag %s does not exists. Creating at commit %s",
+        DOCUMENTATION_TAG,
+        clients.repository.current_commit,
+    )
+    clients.repository.tag_commit(DOCUMENTATION_TAG, clients.repository.current_commit)
+    return True
```
gregory-schiano, line 250: I'm wondering if the logic shouldn't be a bit refactored, something like:
```python
if not documentation_commit:
  logging.info(...)
  clients.repository.tag_commit(...)
  return True
if flag
  return True
logging.error(...)
return False
```

It reduces indentation, manages return True first, then finally the error case.
Ideally the return False should be first, but that'll probably less readable

gregory-schiano: Great, thx :)

gregory-schiano (approved):
LGTM

jdkandersson (approved):

PR #212, opened 2023-09-15
Support external links

jdkandersson:
Applicable spec: https://discourse.charmhub.io/t/specification-isd003-upload-charm-docs-contents-index-and-navigation-table/7986

### Overview

Adds support for external links on the contents index which are then written to the navigation table.

### Rationale

Including external links enables them to be added to the navigation table which is a requirement for documentation.

### Module Changes

* Introduced several new actions for reconcile to distinguish between files, groups and external items
* Introduced attribute to discourse client with the base path to be able to distinguish between topic and external links
* Introduced generation of the new action types in the reconcile and handling of them in the action module

### Checklist

Version should be incremented on release


src/index.py
```diff
@@ -269,35 +304,39 @@ def _check_contents_item(
         )
 
     # Check whether item is hidden and a directory
+    item_reference_type = classify_item_reference(
+        reference=item.reference_value, docs_path=docs_path
+    )
     item_path = docs_path / Path(item.reference_value)
```
gregory-schiano, line 310: Shouldn't this be moved after l314 ?

src/index.py
```diff
@@ -269,35 +304,39 @@ def _check_contents_item(
...
+        if len(item_to_aggregate_path.parents) != 1:
             raise InputError(
-                "An item in the contents list is not of the expected file type. "
-                f"{item=!r}, expected extension: {DOC_FILE_EXTENSION}"
+                "A nested item is a reference to a path that is not immediately within the "
+                f"directory of its parent. {item=!r}, expected parent path: {aggregate_dir!r}"
             )
 
+        # Check that if the item is a file, it has the correct extension
+        if item_path.is_file():
```
gregory-schiano, line 333: Shouldn't we use `item_reference_type == ItemReferenceType.FILE` instead ?

gregory-schiano (commented):
Stopped at `migration.py` the PR is a bit more complex than I expected

src/action.py
```diff
@@ -17,17 +17,33 @@
 NOT_DELETE_REASON = "delete_topics is false"
 
 
-def _absolute_url(url: types_.Url | None, discourse: Discourse) -> types_.Url | None:
+def _absolute_url(
+    action: types_.AnyAction, url: types_.Url | None, discourse: Discourse
```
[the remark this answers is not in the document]
jdkandersson: The `url` doesn't always come from the action, sometimes it comes from the `TableRow` and, depending on the action, the location of the link varies as well

src/index.py
jdkandersson: yes this can be moved now

src/index.py
jdkandersson: thanks, should be fixed now

src/types_.py
```diff
@@ -168,41 +166,93 @@ def is_group(self) -> bool:
...
+
+        Returns:
+            Whether the item in the table is an external item.
+        """
+        if self.navlink.link is None:
+            return False
+        comparison_link = self.navlink.link.lower()
+        return comparison_link.startswith("http") and not comparison_link.startswith(
+            server_hostname.lower()
+        )
```
gregory-schiano, line 183: Couldn't we only rely on `server_hostname` prefix ?
If `server_hostname` doesn't start with `http` it will be considered as external (so we probably need to ensure it starts with it)

gregory-schiano: I'm wondering also if we shouldn't have some more validation on the URL, to avoid some risks, this link will endup on the public page of charmhub. That's probably not there that this validation will occur, but we want to avoid things like `http://test*some markdown or html*`

src/types_.py
gregory-schiano, line 1: For the new dataclasses could be keep the same order in which they are declared, like first the `group`, then `page`, then `externalRef` ?
They're currently each in different order (in create, noop, update, and delete)

src/types_.py
```diff
@@ -444,6 +561,24 @@ class IndexContentsListItem(typing.NamedTuple):
     rank: int
     hidden: bool
 
+    @property
+    def table_path(self) -> TablePath:
+        """The table path for the item.
```
gregory-schiano, line 566: The description should also mention the sanitization applied to the reference_value

tests/integration/test___init__run_migrate.py
```diff
@@ -85,6 +85,7 @@ async def test_run_migrate(
 
     | Level | Path | Navlink |
     | -- | -- | -- |
+    | 1 | https-canonical-com | [Canonical](https://canonical.com) |
```
gregory-schiano, line 88: We probably also should have a case with a more complex url that contains query strings or anchors

gregory-schiano (commented):
Reviewed until the tests

src/reconcile.py
```diff
@@ -116,23 +138,58 @@ def _local_and_server_dir_local_group_server(
...
+            ),
+        )
+    return (
+        types_.UpdateExternalRefAction(
+            level=item_info.hierarchy,
+            path=item_info.table_path,
+            navlink_change=types_.NavlinkChange(
+                old=table_row.navlink,
+                new=types_.Navlink(
+                    title=item_info.reference_title, link=item_info.reference_value, hidden=False
```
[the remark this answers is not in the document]
jdkandersson: Yes that is probably better, there isn't really a use case for hidden external items. It is probably better for consistency and so that the behaviour is not unexpected for users

src/types_.py
jdkandersson: Re validation, there is a further step to check that a HEAD request to the external link works in a future PR which I think will address that need, what do you think?

jdkandersson: The issue regarding just relying on `comparison_link` is that many of the tests don't use a discourse link, they just use `link 1` or something similar. If this line is changed, 50 tests fail. Ideally we would change this so that we get correct discourse links in the factories - the scope of this PR is quite large so I'll add a follow up issue if this is a reasonable path forward

tests/integration/test___init__run_migrate.py
jdkandersson: Since for migration the link is effectively skipped based on the `TableRow.is_external` function, do you think it would be better to add those to test for `is_external`? Anchors and query strings are indeed a good case for which the path creation needs to be updated

src/action.py
```diff
@@ -81,6 +104,9 @@ def _create(
             url = FAIL_NAVLINK_LINK
             result = types_.ActionResult.FAIL
             reason = str(exc)
+    # This should not occur since all the create action types have been implemented
+    else:  # pragma: no cover
+        raise NotImplementedError(f"internal error: {action} has not been implemented")
```
yhaliaw, line 109: I am guessing this is targeting python 3.10 so this does not apply, but they added an [`assert_never` in Python 3.11](https://docs.python.org/3/library/typing.html#typing.assert_never).

src/action.py
```diff
@@ -56,21 +72,28 @@ def _create(
 
     Returns:
         A report on the outcome of executing the action.
+
+    Raises:
+        NotImplementedError: if a requested action has not been implemented yet.
```
yhaliaw, line 77: Since the code is not expecting the error to be ever raised (line 107), should this be in the docstring?

jdkandersson: It gets enforced by our tooling - it also seems good to at least document the possibility that it might be raised

src/action.py
jdkandersson: Yeh this would be interesting - the code is actually compliant with 3.11 (and in production 3.11 is used) - just in `operator-workflows` we're using `22.04` which comes with 3.10 and hence the testing and linting is run on 3.10

src/index.py
```diff
@@ -269,35 +304,39 @@ def _check_contents_item(
...
+    if item.hidden and item_reference_type == ItemReferenceType.DIR:
+        raise InputError(f"A hidden item is a directory. {item=!r}")
 
-    # Check that the item is directly within the current directory
-    if len(item_to_aggregate_path.parents) != 1:
-        raise InputError(
-            "A nested item is a reference to a path that is not immediately within the "
-            f"directory of its parent. {item=!r}, expected parent path: {aggregate_dir!r}"
-        )
+    if item_reference_type in {ItemReferenceType.DIR, ItemReferenceType.FILE}:
```
yanksyoon, line 314: Out of curiosity, may I ask why a set is used rather than a tuple here please?

src/reconcile.py
```diff
@@ -468,17 +665,22 @@ def run(
...
     Args:
         index: Information about the index on the server and locally.
         table_rows: The current navigation table rows based on local files.
+        discourse: A client to the documentation server.
 
     Returns:
         The action to take for the index page.
     """
-    table_contents = "\n".join(table_row.to_markdown() for table_row in table_rows)
+    table_rows = list(table_rows)
```
yanksyoon, line 680: Out of curiosity, why is a list preferred here over tuple (I'm not sure I see any mutation on `table_rows`)?

yanksyoon (commented):
Reviewed until `src/types_.py` (just until tests), a few questions! Thank you! :D

src/action.py
```diff
@@ -299,18 +318,42 @@ def _run_one(
...
             # To help mypy (same for the rest of the asserts), it is ok if the assert does not run
-            assert isinstance(action, types_.CreateAction)  # nosec
+            assert isinstance(
+                action,
+                (
+                    types_.CreatePageAction,
+                    types_.CreateGroupAction,
+                    types_.CreateExternalRefAction,
+                ),
+            )  # nosec
```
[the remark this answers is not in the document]
jdkandersson: yes, that makes it simpler actually, should be fixed shortly

src/index.py
jdkandersson: In practice the difference is negligible - the reason is that checking for whether a value is in a set is `O(1)` and checking whether it is in a tuple is `O(N)`. In this case N is 2 - its just that there isn't really a cost to using a set so why wouldn't a set be used?

src/reconcile.py
jdkandersson: I think this is leftover from debugging - removed the conversion to `list`

yanksyoon (approved):
LGTM!

src/reconcile.py
```diff
@@ -61,27 +76,37 @@ def _get_server_content(table_row: types_.TableRow, discourse: Discourse) -> str
...
+            )
+    if isinstance(item_info, types_.IndexContentsListItem):
+        if item_info.hierarchy != table_row.level:
+            raise exceptions.ReconcilliationError(
+                f"internal error, level mismatch, {item_info=!r}, {table_row=!r}"
+            )
+        if item_info.table_path != table_row.path:
+            raise exceptions.ReconcilliationError(
+                f"internal error, table path mismatch, {item_info=!r}, {table_row=!r}"
+            )
```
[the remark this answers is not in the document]
jdkandersson: Fixed now

src/reconcile.py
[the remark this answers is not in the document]
jdkandersson: Fixed now

src/types_.py
jdkandersson: Done now

src/types_.py
jdkandersson: Done now

tests/integration/test___init__run_migrate.py
jdkandersson: I have added a few more tests to `is_external` and the algorithm for creating the path

deusebio (commented):
I started to go through the PR, but then I started to have some thoughts about what we mean by external/internal link.

I believe Discourse and frontend platforms (e.g. charmhub) allow for different behaviour when one provides 
```
[Link](https://discourse.com/t/my-topic)
```
and
```
[Link](/t/my-topic)
```

The first does represent an actual re-direct to Discourse, the second is a reference which is handled by the frontends (e.g. charmhub) by rendering the other content, still within the platform. 

So my point here is that `https://discourse.com/url` may possibly still be treated as an external link, regardless of whether it starts with the server name or not. This may also allow in the README ( (consistently with the different behaviour in charmhub) to refer to discourse documentation outside of the scope of the product, without copying/rendering the content in Github, e.g. :

```
[Content In Docs]("/tutorial/doc.md") # Link rendered in Github
[Content In Other Docs]("https://dicourse.com/tutorial/doc.md") # Link redirected to Discourse
```

jdkandersson:
> I started to go through the PR, but then I started to have some thoughts about what we mean by external/internal link.
> 
> I believe Discourse and frontend platforms (e.g. charmhub) allow for different behaviour when one provides
> 
> ```
> [Link](https://discourse.com/t/my-topic)
> ```
> 
> and
> 
> ```
> [Link](/t/my-topic)
> ```
> 
> The first does represent an actual re-direct to Discourse, the second is a reference which is handled by the frontends (e.g. charmhub) by rendering the other content, still within the platform.
> 
> So my point here is that `https://discourse.com/url` may possibly still be treated as an external link, regardless of whether it starts with the server name or not. This may also allow in the README ( (consistently with the different behaviour in charmhub) to refer to discourse documentation outside of the scope of the product, without copying/rendering the content in Github, e.g. :
> 
> ```
> [Content In Docs]("/tutorial/doc.md") # Link rendered in Github
> [Content In Other Docs]("https://dicourse.com/tutorial/doc.md") # Link redirected to Discourse
> ```

@deusebio An external link is intended to be something that isn't a link to a topic that should be rendered as documentation. You are right that this could be a discourse topic which would currently get treated as a documentation topic. The current approach to handling URLs is that, during execution, the full URL to the topic is used for discourse interactions. So in many cases the distinction between a discourse topic intended as documentation and an external link to a discourse topic would be lost. This indicates that there is a refactor needed of the URL handling which I would propose for a follow up PR along with cleaning up the URL generation in tests. For now, this adds support for linking, e.g., to the configuration or actions from the navigation table which is commonly done and recommended so it represents progress wghich is why I suggest we proceed with it.

src/discourse.py
```diff
@@ -485,6 +489,11 @@ def update_topic(
 
         return self.absolute_url(url=url)
 
+    @property
+    def base_path(self) -> str:
```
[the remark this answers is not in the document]
jdkandersson: Agree, I'll switch to `base_url`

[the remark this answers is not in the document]
jdkandersson: done

src/types_.py
jdkandersson: yes the `http` prefix should be enough - its just that some refactoring is needed to make that possible

src/types_.py
```diff
@@ -444,6 +564,35 @@ class IndexContentsListItem(typing.NamedTuple):
...
+        case. Any remaining characters not allowed in the path are also removed. For a non-HTTP
+        case, removes the file suffix and splits on / to built the path.
+
+        Returns:
+            The table path for the item.
+        """
+        if self.reference_value.lower().startswith("http"):
+            transformed_reference_value = (
+                self.reference_value.replace("//", "/")
+                .replace(":", "")
```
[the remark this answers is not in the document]
jdkandersson: I think this can be removed now anyway

src/types_.py
deusebio: Yes, this may then simplify quite a bit the code, as also passing down the server_hostname would not be required. Also this seems also something is consistent with another line [here](https://github.com/canonical/upload-charm-docs/blob/4d1d52ba83db9239c24010f5f860699223087f88/src/index.py#L272)

src/types_.py
```diff
@@ -444,6 +564,34 @@ class IndexContentsListItem(typing.NamedTuple):
...
+        Returns:
+            The table path for the item.
+        """
+        if self.reference_value.lower().startswith("http"):
+            transformed_reference_value = (
+                self.reference_value.replace("//", "/")
+                .replace(".", "/")
+                .replace("?", "/")
+                .replace("#", "/")
+            )
```
deusebio, line 585: It is small stuff, but why not using regex then? it seems also more consistent with the next line

deusebio:
@jdkandersson Yes, that's fine. I thought that changing the logic would have been only on the `is_external` function and fairly self-contained and straightforward, but - yet - I believe it is important to define what `external` means. It is fine to approach this in a subsequent PR if this means a substantial rework in this PR, which is more around adding support for external link, which I agree and I believe it is sensible.

deusebio (approved):

src/types_.py
jdkandersson: The string replace is easier to read, the regex replace is intended to capture any other characters that are not allowed on the `path` column - it would be nicer to do that as a string replace its just that the definition of what is allowed on the path column is already defined as a regex

yanksyoon (approved):
LGTM!

PR #217, opened 2023-09-28
Migrate navigation table

jdkandersson:
Applicable spec: https://discourse.charmhub.io/t/specification-isd003-upload-charm-docs-contents-index-and-navigation-table/7986

### Overview

Generates the contents index from the navigation table during migration

### Rationale

so that the navigation table is now also fully managed

### Module Changes

The migrate module has been changed to generate the contents index during migration

### Checklist

Version to be incremented on release to stable


src/migration.py
```diff
@@ -84,7 +97,7 @@ def _create_document_meta(row: types_.TableRow) -> types_.DocumentMeta:
             "Internal error, no implementation for creating document meta with missing link in row."
         )
     return types_.DocumentMeta(
-        path=Path(*row.path[:-1]) / f"{row.path[-1]}.md", link=row.navlink.link, table_row=row
+        path=_generate_document_path(row), link=row.navlink.link, table_row=row
```
weiiwang01, line 100: Could it be `Path(*row.path).with_suffix(".md")`?

src/navigation_table.py
```diff
@@ -31,7 +31,8 @@
 _PATH_REGEX = rf"{_WHITESPACE}([${constants.PATH_CHARS}]+){_WHITESPACE}"
 _PUNCTUATION = string.punctuation.replace("/", "\\/")
 _NAVLINK_TITLE_REGEX = rf"[\w\- {_PUNCTUATION}]+?"
-_NAVLINK_LINK_REGEX = r"[\w\/-]*"
+# Link characters according to https://www.rfc-editor.org/rfc/rfc3986#section-2
+_NAVLINK_LINK_REGEX = r"[\w\/._~:/?#\[\]@!$&'*+,;-]*"
```
weiiwang01, line 35: I think rfc3986 defines `"(" | ")"` to be valid characters in the URI, but URI containing brackets might just be a very edge case so we can ignore that.

jdkandersson: yeh I removed those because it would interfere with the markdown link which puts brackets around the link

src/migration.py
weiiwang01: Sorry I think the `with_suffix` will replace the content after the dot, so a `1. Introduction` will be converted to `1.md`. The current approach is probably safer.

weiiwang01 (approved):

merkata (approved):

PR #231, opened 2024-01-18
Fix nothing to migrate bug

jdkandersson:
Applicable spec: N/A

### Overview

There was an issue where the action attempts to perform migration when the default branch and discourse match but the base content tag is behind. This now no longer attempts to perform migration and also moves the tag to the latest commit on the default branch.

### Rationale

To resolve a bug as described above.

### Module Changes

The migration module no longer attempts to migrate when the default branch and discourse are up to date.

### Checklist

- [x] Version has been incremented


src/__init__.py
```diff
@@ -221,11 +221,19 @@ def run_migrate(clients: Clients, user_inputs: UserInputs) -> MigrateOutputs | N
...
     changes = recreate_docs(clients, DOCUMENTATION_TAG)
+    # Check whether there are still changes after switching to the base branch
+    if changes:
+        changes = clients.repository.is_dirty(user_inputs.base_branch)
+
+        # Move the tag if there are no changes
+        if not changes:
+            with clients.repository.with_branch(user_inputs.base_branch) as repo:
+                main_hash = repo.current_commit
+            clients.repository.tag_commit(DOCUMENTATION_TAG, main_hash)
```
deusebio, line 232: **suggestion** maybe it is not needed to defined `main_hash`. Would this work?

```
with clients.repository.with_branch(user_inputs.base_branch) as repo:
    clients.repository.tag_commit(DOCUMENTATION_TAG, repo.current_commit)
```

deusebio (approved):
Changes look fine!

src/__init__.py
jdkandersson: I think it is required because otherwise the branch at this point isn't the base branch. I tried removing it and the wrong commit SHA is tagged

jdkandersson: On line 218 it is done in the same way

cbartz (dismissed):
LGTM

src/__init__.py
```diff
@@ -221,11 +221,19 @@ def run_migrate(clients: Clients, user_inputs: UserInputs) -> MigrateOutputs | N
...
+
     if not changes:
         logging.info(
-            "No community contribution found in commit %s. Discourse is inline with %s",
-            user_inputs.commit_sha,
-            DOCUMENTATION_TAG,
+            "No community contribution found, discourse is inline with %s", DOCUMENTATION_TAG
```
yanksyoon, line 236: can this be merged under `if not changes:` line above?

yanksyoon (approved):
LGTM

src/__init__.py
jdkandersson: The `if not` above is in another `if` block, or do you mean another location?

deusebio (approved):

PR #238, opened 2024-03-22
Prepare discourse-gatekeeper for PAAS app charmer 

javierdelapuente:
Applicable spec: <link>

### Overview

These are the requirements for PAAS app charmer with respect to discourse-gatekeeper:
 - Use charmcraft.yaml as there is no metadata.yaml. 
 - Use a custom directory instead of the root directory for the charm. Also, the docs will be inside the charm directory.

Currently, the name and the link to the documentation are just read from the `metadata.yaml` file. However, the new recommended way is to set in the `charmcraft.yaml` file (see https://juju.is/docs/sdk/charmcraft-yaml). This PR reads those fields from `charmcraft.yaml` if `metadata.yaml` does not exist. This PR does not take into account the case where name is one file and the link to the doc is in the other file. In that case, the suggestion could be to put both fields in the same file (preferably charmcraft.yaml).

Besides there is a new input variable, charm_dir, thats allows to customise the place where the metadata/charmcraft file is placed and where the docs directory is.

It was necessary to delete the `rate_limit_new_user_create_topic`, as it does not exist anymore in discourse (before that, it was not used in discourse anyway).

### Rationale

It was not possible to use this project for PAAS app charmer autogenerated projects, as they use the `charmcraft.yaml` file and the charm directory is not in the root of the repository.

### Checklist

- [x] Version has been incremented

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Added a formatted representation of the UpdatePageAction dataclass for more
   human-readable output.
-- If metadata.yaml does not exist, is reads name and doc information from
+- If metadata.yaml does not exist, read the name and doc information from the
   charmcraft.yaml file.
 - New input environment variable INPUT_CHARM_DIR. metadata.yaml or charmcraft.yaml
   will be read from this directory instead of the base one and the documentation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,14 @@
 # Changelog
 
-## [Unreleased]
+## [v0.8.3] - 2024-03-26
 
 - Added a formatted representation of the UpdatePageAction dataclass for more
   human-readable output.
+- If metadata.yaml does not exist, is reads name and doc information from
+  charmcraft.yaml file.
+- New input environment variable INPUT_CHARM_DIR. metadata.yaml or charmcraft.yaml
+  will be read from this directory instead of the base one and the documentation
+  will also be searched under this directory.
 
 ## [v0.8.2] - 2024-02-16
 
--- a/action.yaml
+++ b/action.yaml
@@ -55,6 +55,14 @@ inputs:
       variable GITHUB_SHA.
     required: false
     type: string
+  charm_dir:
+    description: |
+      Relative name of the directory where the metadata.yaml or charmcraft.yaml files are located
+      if they are not in the root directory of the repository. The docs directory is also located
+      under this directory.
+    default: ''
+    required: false
+    type: string
 outputs:
   index_url:
     description: |
--- a/src/metadata.py
+++ b/src/metadata.py
@@ -3,8 +3,6 @@
 
 """Module for parsing metadata.yaml file."""
 
-import logging
-import os
 from pathlib import Path
 
 import yaml
--- a/src/metadata.py
+++ b/src/metadata.py
@@ -39,13 +39,6 @@ def get(path: Path) -> types_.Metadata:
         InputError: if the metadata file does not exist or are malformed.
 
     """
-    logging.info("metadata.get, path: %s abs path: %s", path, path.absolute())
-    logging.info("metadata.get, cwd %s", os.getcwd())
-    logging.info("metadata.get, list dir cwd %s", os.listdir())
-    logging.info("metadata.get, list dir cwd .. %s", os.listdir(".."))
-    logging.info("metadata.get, list dir sent path. %s", os.listdir(path))
-    logging.info("metadata.get, list dir sent path .. %s", os.listdir(path / ".."))
-
     metadata_yaml = path / METADATA_FILENAME
     if metadata_yaml.is_file():
         return _parse_metadata_yaml(metadata_yaml)
--- a/src/metadata.py
+++ b/src/metadata.py
@@ -34,7 +34,7 @@ def get(path: Path) -> types_.Metadata:
         The contents of the metadata file.
 
     Raises:
-        InputError: if the metadata file does not exist or are malformed.
+        InputError: if the metadata file does not exist or is malformed.
 
     """
     metadata_yaml = path / METADATA_FILENAME
--- a/src/metadata.py
+++ b/src/metadata.py
@@ -4,6 +4,7 @@
 """Module for parsing metadata.yaml file."""
 
 from pathlib import Path
+import os
 import logging
 
 import yaml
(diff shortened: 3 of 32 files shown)


arturo-seijas (dismissed):

CHANGELOG.md
```diff
@@ -1,9 +1,14 @@
 # Changelog
 
-## [Unreleased]
+## [v0.8.3] - 2024-03-26
 
 - Added a formatted representation of the UpdatePageAction dataclass for more
   human-readable output.
+- If metadata.yaml does not exist, is reads name and doc information from
```
jdkandersson, line 7: Perhaps `If metadata.yaml does not exist, read the name and doc information from the charmcraft.yaml file.`

action.yaml
```diff
@@ -55,6 +55,14 @@ inputs:
...
     required: false
     type: string
+  charm_dir:
+    description: |
+      Relative name of the directory where the metadata.yaml or charmcraft.yaml files are located
+      if they are not in the root directory of the repository. The docs directory is also located
+      under this directory.
+    default: ''
+    required: false
+    type: string
```
jdkandersson, line 65: We should probably also add an e2e test for this here for the new input please: https://github.com/canonical/gatekeeper-repo-test

src/repository.py
```diff
@@ -248,7 +272,7 @@ def with_branch(self, branch_name: str) -> Iterator["Client"]:
         finally:
             self.switch(current_branch)
 
-    def get_summary(self, directory: str | None = DOCUMENTATION_FOLDER_NAME) -> DiffSummary:
+    def get_summary(self, directory: str | None = "") -> DiffSummary:
```
jdkandersson, line 275: Since we default `directory` at the GitHub level it can be dangerous to spread these kind of defaults throughout the repo since if they change in one place they have to change in all places. I'm guessing this is done to reduce the need to update tests? It would be good to remove the default and rely on the GitHub input default. Any thoughts on that please?

jdkandersson: Also applicable to any other places where we have defined the default

tests/integration/conftest.py
jdkandersson, line 1: Do you think we should have an integration test for the new user input?

tests/unit/helpers.py
jdkandersson, line 1: Not exactly related to this file, there are a few functions that now accept a new argument around the changed directory. Should we add unit tests with non-default values for those functions?

tests/unit/test_metadata.py
```diff
@@ -124,3 +125,109 @@ def test_get(
...
+    create_charmcraft_yaml(content=charmcraft_yaml_content, path=tmp_path)
+
+    with pytest.raises(exceptions.InputError) as exc_info:
+        metadata.get(path=tmp_path)
+
+    assert_substrings_in_string(expected_error_msg_contents, str(exc_info.value).lower())
+
+
+# pylint currently does not understand walrus operator perfectly yet
+# pylint: disable=unused-variable,undefined-variable
```
jdkandersson, line 197: I know this is not easy to know, there is actually an alternative pattern for how to do this which has partially been implemented and we have been migrating tests to that pattern over time. See here for an example: https://github.com/canonical/discourse-gatekeeper/blob/5c6f9bedadf2137b9e6f93ed2b50fcdb58fca084/tests/unit/test_check.py#L17

jdkandersson (commented):
Thanks for creating this, it has been in the back of my mind to do this for a while, just haven't found the time

tests/unit/test_metadata.py
jdkandersson: It would be great if you could update the all the tests at least in this file to use the new pattern, no need to do them all, just keep making progress 🙂

javierdelapuente: sure, I will change that file. I have already seen that way, but I did not know that was the recommended way. TBH I found the walrus operator a bit smart but nice!

src/repository.py
javierdelapuente: Done!

tests/unit/test_metadata.py
javierdelapuente: Done!

src/metadata.py
```diff
@@ -62,3 +93,55 @@ def get(path: Path) -> types_.Metadata:
...
+        charmcraft_yaml: The file path the the charmcraft file.
+
+    Returns:
+        The contents of the charmcraft file.
+
+    Raises:
+        InputError: if the charmcraft file does not exist or are malformed.
+    """
+    try:
+        charmcraft = yaml.safe_load(charmcraft_yaml.read_text())
```
amandahla, line 111: Minor: Could a FileNotFoundError happen here?

amandahla (approved):
Cool! 
Left a minor question.

tests/integration/test___init__run_reconcile.py
```diff
@@ -29,12 +29,21 @@
 
 
 @pytest.mark.asyncio
-@pytest.mark.usefixtures("patch_create_repository_client")
+@pytest.mark.usefixtures("git_repo")
+@pytest.mark.parametrize(
+    "charm_id, charm_dir",
+    [
+        ("1", ""),
+        ("2", "charm"),
```
jdkandersson, line 37: I think if you remove the trailing comma here it might put more of this on the same line

jdkandersson (commented):
Looks good, looking forward to the integration and e2e tests

src/metadata.py
javierdelapuente: in the get function that calls this one, the file is checked, so it should exist by this point.

tests/integration/conftest.py
javierdelapuente: Updated so the two main tests run twice, with the base path and with another path.

tests/unit/helpers.py
javierdelapuente: I have created another unit test for the function with non-default value. There are a few other refactoring in the tests that could be nice, but I believe it would be better to do it in another PR (like accessing directly the docs directory).

amandahla (approved):
LGTM

jdkandersson (approved):

action.yaml
jdkandersson: Let's try and get it working, just please include a link here to a successful run on a branch and then we can merge this before the change to the e2e test

javierdelapuente: There are e2e tests that test the new functionality in here:
- https://github.com/canonical/gatekeeper-repo-test/actions/runs/8505195956
- https://github.com/canonical/gatekeeper-repo-test/actions/runs/8505195954

PR #244, opened 2024-04-09
Upgrade project structure so a package can be build

javierdelapuente:
Applicable spec: <link>

### Overview

Currently, the repo https://github.com/canonical/gatekeeper-repo-test has a copy of all the code of this (discourse-gatekeeper) repo.

This is problematic, as it has to be updated from time to time. This PR prepares discourse-gatekeeper to be used as a package, so it can be easily included in https://github.com/canonical/gatekeeper-repo-test (see https://github.com/canonical/gatekeeper-repo-test/pull/466 for a working example).

The main change of this PR is to put all the code under the gatekeeper module (instead of just under src) and preparing pyproject.toml for the building.

### Checklist


deusebio (approved):
This is REALLY great! It was definitely needed. I'm still ashamed of myself for doing the copy and paste of the source code for the discourse-test :joy:

deusebio:
Actually, one thought: maybe it would make things a bit more structured (and it should be rather straightforward) if when we merge to main, we also release the python package with github releases. Then one can simply install this using

```
pip install https://github.com/canonical/discourse-gatekkeeper/releases/download/vX.Y.Z/gatekeeper-X.Y.Z-py3-none-any.whl
```

Eventually we could publish this to pypi, but it is probably not needed right now, and I believe github releases would be good enough. This would however provide a bit of better versioning of this and allow version pinning. I would really not use 

```
git+https://github.com/canonical/discourse-gatekeeper.git@main
```

We have done these things above in [this repository](https://github.com/canonical/spark-k8s-toolkit-py): [here is the CI](https://github.com/canonical/spark-k8s-toolkit-py/blob/main/.github/workflows/release_github.yaml#L60) and [here are the releases](https://github.com/canonical/spark-k8s-toolkit-py/releases/tag/v0.0.6).

.dockerignore
```diff
@@ -5,3 +5,4 @@
 !/main.py
 !/requirements.txt
 !/src
+!/pyproject.toml
```
jdkandersson, line 8: Perhaps a final new line?

jdkandersson (approved):

jdkandersson:
> Actually, one thought: maybe it would make things a bit more structured (and it should be rather straightforward) if when we merge to main, we also release the python package with github releases. Then one can simply install this using
> 
> ```
> pip install https://github.com/canonical/discourse-gatekkeeper/releases/download/vX.Y.Z/gatekeeper-X.Y.Z-py3-none-any.whl
> ```
> 
> Eventually we could publish this to pypi, but it is probably not needed right now, and I believe github releases would be good enough. This would however provide a bit of better versioning of this and allow version pinning. I would really not use
> 
> ```
> git+https://github.com/canonical/discourse-gatekeeper.git@main
> ```
> 
> We have done these things above in [this repository](https://github.com/canonical/spark-k8s-toolkit-py): [here is the CI](https://github.com/canonical/spark-k8s-toolkit-py/blob/main/.github/workflows/release_github.yaml#L60) and [here are the releases](https://github.com/canonical/spark-k8s-toolkit-py/releases/tag/v0.0.6).

I agree with this, probably we should do it as well :)

jdkandersson (approved):

deusebio:
> Do you mind to leave it for a following PR? Its first use is just for the e2e tests, which will use branches (and using main and failing is also good, as it is a test). Clearly publishing is better for the main case if the library is used outside the e2e use.

It was not a blocking comment on my side (actually I had already approved I believe). Up to you guys

deusebio (approved):

Issue #249, opened 2024-06-17
Issues with ToC parsing and importing pages. 


a-velasco:
Submitting this issue report as a log of unexpected behavior that might be related to this [existing issue](https://github.com/canonical/discourse-gatekeeper/issues/218) .    

## Context
Several Data Platform charms have been using discourse-gatekeeper for a one-way sync. We recently updated one of our repos ([`mysql-router-k8s-operator`](https://github.com/canonical/mysql-router-k8s-operator) | [Charmed MySQL Router K8s](https://charmhub.io/mysql-router-k8s)) to use `canonical/discourse-gatekeeper@stable` instead of a [fork](deusebio/discourse-gatekeeper@c8adb89ea1cbceca54d78da798658373615487ac) that we've been using.

The [first run](https://github.com/canonical/mysql-router-k8s-operator/actions/runs/9545477480/job/26306422060) of the updated action passed, the only difference being that it generated a new table of contents in the index:
```diff
+# Contents
+
+1. [Tutorial](tutorial)
+  1. [1. Introduction](tutorial/t-overview.md)
+  1. [2. Set up the environment](tutorial/t-setup-environment.md)
...
```
Full diff: ([PR 256](https://github.com/canonical/mysql-router-k8s-operator/pull/256/files)) 

## Issues
1. The action does not generate or parse the table of contents correctly.
2. Running the action on a repo with no `/docs` folder creates a complete table of contents in `index.md`, but only imports some of those pages.

## Steps taken
**After the initial update, we triggered the workflow with no changes to discourse or GitHub.** 
It failed due to an index/ToC related error ([link to logs](https://github.com/canonical/mysql-router-k8s-operator/actions/runs/9545585587/job/26306648411)):
```shell
src.exceptions.InputError: A nested item is a reference to a path that is not within the directory of its parent. item=_ParsedListItem(whitespace_count=0, reference_title='Reference', reference_value='reference', rank=19, hidden=False), expected parent path: PosixPath('how-to')
```

**We switched the discourse-gatekeeper call to `main`** in case there was a related fix not yet merged into `stable` - but got the same error.

**Then, we deleted the entire `/docs` directory to allow the action to re-generate all docs from scratch** with the expected structure. However, the action only downloaded a select few pages.
[This older revision](https://github.com/canonical/mysql-router-k8s-operator/tree/1a28465fcd3993d69aaae8999945f1a7085350a8/docs) shows the expected pages.
[...]

PR #250, opened 2024-06-27
Extend hierarchy calculation for deeply nested pages

a-velasco:
### Overview

A small change to the calculation of page hierarchy to account for navigation tables with large jumps in nesting. 

Previously, the difference in whitespace only got evaluated once, which only accounts for cases where the next item is nested one level higher or lower - but not more.

### Rationale

This change is needed because the action fails on documentation sets with hierarchy jumps greater than 1. More context + examples in issue #249.   

### Module Changes

None

### Checklist

- [ ] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,7 @@
 
 - Project prepared to be installed with `pip install`, so it can be reused in 
   the repository https://github.com/canonical/gatekeeper-repo-test
-
-## [v0.9.1] - 2024-06-28
-
-### Fixed
-
-- Support for navigation table with hierarchy jumps of more than 1 level
+- Added support for navigation table with hierarchy jumps of more than 1 level
 
 ## [v0.9.0] - 2024-04-04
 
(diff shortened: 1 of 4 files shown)


jdkandersson (dismissed):

jdkandersson:
/canonical/self-hosted-runners/run-workflows 1d27df709c2e16cb56f91310322eac572c1328af

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@
 - Project prepared to be installed with `pip install`, so it can be reused in 
   the repository https://github.com/canonical/gatekeeper-repo-test
 
+## [v0.9.1] - 2024-06-28
+
+### Fixed
```
jdkandersson, line 10: Until released it should be noted under `Unreleased` (i.e., without a date heading) please. Once we make a release with this, it'll get a date and version

jdkandersson:
It also looks like there have been changed to the discourse change, the ad admin user is no longer there (required for testing), it is now create user ands then promote user. Would you like to fix that?

